PR #11, opened 2015-11-04
Mark backwards compat modules as not going to be ported.

abadger:
These packages are backwards compat packages.  They're older versions of an upstream package that we created just for an application that hadn't yet ported to a newer version.

The older version will never be ported to python3 -- for these particular packages, using the current version of the package is the correct porting strategy as upstream for these packages has included python3 support in the current version of the package (and those versions are packaged in Fedora).


encukou:
Thanks! :balloon:

PR #14, opened 2015-11-13
Just a few more manual notes updates

abadger:
* Upstream notes for tracer and python-lzo
* Remove babel and libxml2 as they're now fixed in rawhide


abadger:
python-mutagen can also be taken out of the manual notes if you update the ```fedora.json``` datafile.

encukou:
Thanks!
I haven't updated now because the 3.5 result is throwing the results off.

PR #19, opened 2015-11-14
Add more upstream data for other packages maintained by the Obnam upstream

michel-slm:
While the packages that Obnam depends on have been marked, there are other related packages by Lars Wirzenius; they should inherit the Obnam upstream status too.

[Review on Reviewable](https://reviewable.io/reviews/fedora-python/portingdb/19)


sYnfo:
LGTM :+1:

PR #37, opened 2015-11-14
Marking python-atfork as in progress

fitoria:


abadger:
Rebased and added here: https://github.com/fedora-python/portingdb/commit/84e4ebae82fa462197100931ee80d1280588270a

PR #38, opened 2015-11-14
Mark argcomplete as done

keszybz:


keszybz:
Let's wait for automatic regeneration instead.

PR #41, opened 2015-11-14
python-flask-debugtoolbar added

frafra:


abadger:
@frafra looks like you pasted the wrong link...  https://bugzilla.redhat.com/show_bug.cgi?id=1282069 goes to the python-printr bug.

frafra:
@abadger Thank you! Now it should be ok.

PR #50, opened 2015-11-14
Updated wiki2beamer

sdyroff:


sdyroff:
No need to add, this info, it is collected automatically

PR #54, opened 2015-11-14
Doing stuff

Riamse:
Stuff is being done


encukou:
Can you add a note to that? Who is doing the work? Is there an repo for the port?

Riamse:
I added a note. I'm confused and I don't think there's a repo for the port. Unless https://github.com/Riamse/keybinder counts as one.

Riamse:
Things are exploding

Riamse:
Screw it I'll have to do this over

PR #57, opened 2015-11-14
Adds information for django-annoying, django-avatar, django-celery

fitoria:


encukou:
Thanks for the contributions!
However, you might be going into mass bug filing territory (see [Fedora wiki](https://fedoraproject.org/wiki/Mass_bug_filing)). We'll need to go through some extra coordination before starting to file bugs for every project that's ported upstream.

PR #61, opened 2015-11-15
Add notes that upstream fail2ban already supports Python3

dperson:


dperson:
This was reported by someone else while I was looking into it.

PR #65, opened 2015-11-15
Add kernel-tests to database

jflory7:
Based on discussions in #fedora-python.

### DIFF
--- a/data/fedora-update.yaml
+++ b/data/fedora-update.yaml
@@ -33,7 +33,7 @@ python-kajiki:
     note: |
         [threebean](http://threebean.org) did this one.  it is built in rawhide.
 kernel-tests:
-    status:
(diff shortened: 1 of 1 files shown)


data/fedora-update.yaml
```diff
@@ -32,6 +32,13 @@ python-kajiki:
     status: released
     note: |
         [threebean](http://threebean.org) did this one.  it is built in rawhide.
+kernel-tests:
+    status:
```
ralphbean: There needs to be some status here, like "in-progress", I think.

jflory7:
@ralphbean Added the status of porting the scripts over. Not sure how I missed that.

ralphbean:
All good.  Thanks @jflory7!

PR #67, opened 2015-11-15
info updated about pylibmc

staranjeet:
The upstream supports Python 3. I have checked the package by running tests. Fedora needs to be updated about this package


sYnfo:
Thanks! :)

PR #74, opened 2015-11-15
Add jabber-roster

fabaff:


sYnfo:
I think it's better to keep it blocked, until the deps are resolved.

fabaff:
Sure

encukou:
Added the info as 68501eb. Thanks!

PR #77, opened 2015-11-15
libssh2-python added

frafra:
I have no real word code to test it, but it seems to work: there's no Python code in it, it's just a binding.


sYnfo:
:+1:
I'd vote for renaming it.

PR #78, opened 2015-11-15
python-bash8 added

frafra:
It produces python2-bash8 and python3-bash8. It's quite strange that python2-bash8 requires both python(abi) = 2.7 and /usr/bin/python3.

### DIFF
--- a/data/fedora-update.yaml
+++ b/data/fedora-update.yaml
@@ -120,6 +120,8 @@ python-backports-ssl_match_hostname:
     status: dropped
     note: |
         Included in Python 3.2
+python-bash8:
+    status: released
 python-BeautifulSoup:
     status: dropped
     note: |


sYnfo:
That's because /usr/bin/bash8 contains #!/usr/bin/python3, the shebang should point to python2 :) I'm gonna leave this PR open until it is fixed.

frafra:
Ok, ok!
I filled a bug report: https://bugzilla.redhat.com/show_bug.cgi?id=1282211

sYnfo:
Wait, no, the executable should even be in the python2 subpackage, it should be in python3-bash8. And it seems to be fixed in rawhide already.

frafra:
Oh, you're right, sorry, it look like this was fixed in v0.1.1-6 (pushed yesterday), while http://portingdb-encukou.rhcloud.com/pkg/python-bash8/ show v0.1.1-4 dependencies.

http://pkgs.fedoraproject.org/cgit/python-bash8.git/commit/python-bash8.spec?id=f3fb90af8e9dfb327354d01529ad996b0ac26d44

Closing bug.

encukou:
This was just picked up by the update script; no need to override the status in the YAML file.

There's a bit of delay since the updates need to get out to mirrors to be picked up.

PR #85, opened 2015-11-15
update python-HTMLgen status

michaeleekk:


encukou:
Are you sure that is a good replacement, not just a project with the same name? The Github repo was last changed 3 months ago, and says *"This library is currently in development. The API is subject to change. Feedback and suggestions are welcome!"*

encukou:
No activity here, on the Fedora bug, or even at https://github.com/srittau/python-htmlgen, so I'll close this.

PR #89, opened 2015-11-15
update pypdf status

michaeleekk:


encukou:
Thanks!

Since pyPdf needs to be replaced in the packages that depend on it, I've set its status to `dropped`.

PR #90, opened 2015-11-15
anjuta package ready in rawhide

MSK61:
anjuta updated and built in rawhide


encukou:
Hi,
Setting the "released" status is automated, by fetching the packages from Rawhide. There's a bit of delay in this (the update needs to go out to mirrors), but by not having the status overriden to "released", any future regression can be picked up.

So, I've just removed the `in-progress` status.

PR #91, opened 2015-11-15
Update status of todocli and python-xlwt

williamjmorenor:


QuLogic:
Information on iotop is already in master.

PR #93, opened 2015-11-15
todocli

williamjmorenor:


encukou:
Thanks for the update!

There's an automated process to pick up these changes, but due to the 3.5 rebuild in Rawhide it's misbehaving a bit. I've run the update more manually; todocli is marked as released now.

PR #98, opened 2015-11-16
Update upstream status of python-rows

williamjmorenor:


encukou:
Thanks!

williamjmorenor:
I know than the Python 3 portin FAD is over but I will keep seenging pull request if they are still usefull

encukou:
With less people, there's less need to coordinate, but it's still useful to know what's being ported :)

Issue #56, opened 2015-11-14
Get "in-progress" statuses for Fedora from Bugzilla


encukou:
This project recently got a lot of contributors, so I'll write down my plans to reduce the chance of duplicate work :)

I should get to these next week (starting 2015-11-16):
* [x] Go through the mass bug-filing procedure to file Fedora bugs for unported packages
* [x] Create a tracking bug for py3 porting
* [x] Add all individual bugs to the tracking bug
* [x] Start filing bugs for individual packages as needed

PR #102, opened 2015-11-16
many packages with python3 support upstream but with outdated specs in Fedora

williamjmorenor:


encukou:
Thanks, pushed.

I did undo the "released" status in fedora-update, though. Setting that is automated by fetching the packages from Rawhide. There's a bit of delay in this (the update needs to go out to mirrors), but by not having the status overriden to "released", any future regression can be picked up.

williamjmorenor:
Thanks for the merge, many of the bugs I filled are being closed, It is usefull if I keep tracking the updates of those packages here? also I need to send patches to maintainers

encukou:
No, you don't need to track the bugs here – when py3 support is released (after it shows up in Rawhide mirrors), the packages will be marked "released" semi-automatically.
(And more automation is coming in #56)

PR #104, opened 2015-11-18
python-xlib added

frafra:
We are discussing how to create an official port to Python 3 on the mailing list.
There are some old patches and an unofficial Python 3 only port.

Here's my full report: https://titanpad.com/ATHhAiShv8


encukou:
Thanks!

PR #107, opened 2015-11-25
Shade graph nodes based on their importance

sYnfo:
IMHO this makes it a little easier to figure out what the most blocking packages are. You currently do lose the blocked/ready-to-port information thought.

Not a final code, just wondering what you think about the idea.

### DIFF
--- a/portingdb/htmlreport.py
+++ b/portingdb/htmlreport.py
@@ -238,7 +238,8 @@ def graph_json():
     query = query.filter(tables.Package.status != 'released')
     query = query.filter(tables.Package.status != 'dropped')
     packages = list(query)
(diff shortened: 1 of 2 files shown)


encukou:
Modify the graph as you like. If you find this easter egg useful, let's also link to it from the homepage :)

You could do the shading in Python; for example [colorsys](https://docs.python.org/3/library/colorsys.html) can be used for darkening (rgb_to_hls, modify the lightness, hls_to_rgb). But Javascript is fine too.

tomspur:
Looks nice. With opacity both informations are still there, e.g. with `.style("fill-opacity", function(d) { return (3.0-d.pending_requirers)/3.0; })`:
screenshot from 2015-11-25 15-08-37

encukou:
This has been live for some time

PR #108, opened 2015-11-25
Update status from iotop, autotrash, powerline and python-ZConfig

williamjmorenor:
Packages now support python3 in rawhide

https://bugzilla.redhat.com/show_bug.cgi?id=1282146

### DIFF
--- a/data/fedora-update.yaml
+++ b/data/fedora-update.yaml
@@ -7,6 +7,9 @@
 anjuta:
     links:
         bug: https://bugzilla.redhat.com/show_bug.cgi?id=1282116
+autotrash:
+    links:
+        bug: https://bugzilla.redhat.com/show_bug.cgi?id=1282044
 boost:
     nonblocking: true
     note: |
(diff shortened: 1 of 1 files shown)


encukou:
Thanks!

I'm almost ready to fetch bugs from a Bugzilla tracker bug (#56), so hopefully this is the last time you need to add these manually.

williamjmorenor:
I have just filled 2 bugs here:

https://bugzilla.redhat.com/show_bug.cgi?id=1282146

So I must wait until the mass bug filling tool or it is usefull to keep filling bugs againg packages with outdated specs, than do not provides a python2 subpackage or with upstream suportting python3 not packaging in Fedora?

encukou:
It's definitely useful.

For the mass filing, we have two options: Use the existing tracker bug, or file a new one. The advantage of filing a new one is that it can have a proper description, so I'd rather do that. But, even then I'll go through the bugs on the existing tracker and assign them to the new one.

PR #109, opened 2015-12-06
Note abiword upstream is already Python 3 compatible.

dhgutteridge:
I took a look at abiword (since it was near the top of the list, and I use it), and there's nothing to do at their end, so I've noted that. I'll open a bug in Red Hat's Bugzilla to move the python-abiword package to Python 3.


dhgutteridge:
Actually, there is something to do at their end, there was something hiding in a Makefile.am file, so I need to revise this.

PR #110, opened 2015-12-06
Note abiword upstream status

dhgutteridge:
I had a look at abiword (since it was near the top of the list, and I use it), so I've noted that. I've also opened a bug in Red Hat's Bugzilla to move the python-abiword package to Python 3. (The only porting issue can be fixed with a trivial patch downstream if need be.)


encukou:
Thank you!

PR #111, opened 2015-12-08
Further updates for upstream status

dhgutteridge:
Abiword applied the patch I sent them, and I also had a look at some other projects I've made use of (diffuse, libxslt, meld, and python-Levenshtein).


encukou:
Thank you for the help!

PR #113, opened 2015-12-14
Unported Turbogears and TurboGears2 dependencies

abadger:
@encukou @ralphbean Updated the fedora and upstream yaml files with information on unported packages in the TG1 and TG2 stack.  Aiming for TG1 to be dropped and TG2 to be updated to have python3 support.


encukou:
Thank you!

PR #116, opened 2015-12-19
Add pointer to sample HTML report server

rodrigc:


encukou:
That should make it easier to find, thanks! I've just changed it to the shorter alias, http://fedora.portingdb.xyz

PR #117, opened 2015-12-19
Update status of Fabric

rodrigc:


encukou:
Thank you!

This is not Fedora-specific, so I've added it to the `porters` file instead.

PR #118, opened 2015-12-30
Add details re. alacarte, python-CDDB, solfege, and wxPython.

dhgutteridge:


encukou:
Thank you!

I've replaced the wxPython info with links to their [py3-compatible implementation](https://github.com/wxWidgets/Phoenix) (which, alas, is not easy to find).

dhgutteridge:
My apologies, I didn't know that existed. (One would think their base site would make a prominent reference to it somewhere...)

PR #122, opened 2016-02-26
Added a few dropped packages and a few packages with a new tag 'dead-upstream'.

torsava:


encukou:
Thanks!
Merged as e6fb5d03e357b0bce4707183259aa41376c62765

PR #123, opened 2016-03-09
Python2-only packages with known python3 support upstream are now labeled 'mispackaged'.

torsava:
Includes an updated fedora.json file.


encukou:
Thanks!

PR #125, opened 2016-03-10
Added dead-upstream and p3 support upsteram notes.

torsava:


encukou:
Instead of python3-pull-request, please set links[bug] in `upstream.yaml`

encukou:
Thanks!
Squashed and merged as a1bd128fe411734cd5739f77b68f1968f74e8936

PR #127, opened 2016-03-15
Update Upstream data

torsava:
Relying from JBeran. The packages are blocked, so I can't file bugs for them (yet).


encukou:
Thanks!
Fixed YAML syntax and merged as 4b3ba725587d13d6589f77777ae8980d029e2a03

PR #130, opened 2016-03-30
Added a new Python 3 Upstream Porting tracking bug.

torsava:
Added a new Python 3 Upstream Porting tracking bug.
Also added a list of tracking bugs that track packages that should be labeled as mispackaged.

fedora.json was updated before and after to make the effected changes observable.


encukou:
Merged as 9c187d5. Thanks!

PR #131, opened 2016-03-30
Noted dead-upstreams, dropped packages and upstream progress.

torsava:
From Dominika and Jan.

### DIFF
--- a/data/upstream.yaml
+++ b/data/upstream.yaml
@@ -151,7 +151,6 @@ fail2ban:
   note: |
       Upstream claims support of python-3.2+
 fife:
-    status: in-progress
     links:
         bug: https://github.com/fifengine/fifengine/issues/631
 fonts-tweak-tool:
@@ -179,7 +178,6 @@ gnome-tweak-tool:
   status: idle
   note: Python 3 is not currently supported upstream.
 gyp:
-    status: in-progress
     links:
         bug: https://codereview.chromium.org/1454433002
 iotop:
@@ -200,9 +198,12 @@ jabber-roster:
   note: |
     Porting will start if deps are available.
 konversation:
-    status: in-progress
     links:
         homepage: https://konversation.kde.org/
+    note:
+        Partially ported: the
+        [media script](https://quickgit.kde.org/?p=konversation.git&a=commit&h=285fc5f496363937b3676f33e7b468b8d85cce5e)
+        is ported to Python 3 while the rest of the package is not.
 libldb:
   status: released
   notes: |
--- a/data/upstream.yaml
+++ b/data/upstream.yaml
@@ -150,6 +150,10 @@ fail2ban:
       bug: https://bugzilla.redhat.com/show_bug.cgi?id=1282498
   note: |
       Upstream claims support of python-3.2+
+fife:
+    status: in-progress
+    links:
+        bug: https://github.com/fifengine/fifengine/issues/631
 fonts-tweak-tool:
   status: released
   links:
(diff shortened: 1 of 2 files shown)


data/upstream.yaml
```diff
@@ -150,6 +150,10 @@ fail2ban:
       bug: https://bugzilla.redhat.com/show_bug.cgi?id=1282498
   note: |
       Upstream claims support of python-3.2+
+fife:
+    status: in-progress
+    links:
+        bug: https://github.com/fifengine/fifengine/issues/631
```
encukou: I don't think this is in progress – there's an issue reported, but no one seems to be working on this.

data/upstream.yaml
```diff
@@ -167,9 +171,17 @@ glances:
...
+    status: in-progress
+    links:
+        bug: https://github.com/basak/glacier-cli/pull/54
 gnome-tweak-tool:
   status: idle
   note: Python 3 is not currently supported upstream.
+gyp:
+    status: in-progress
+    links:
+        bug: https://codereview.chromium.org/1454433002
```
encukou: The review here has been stuck for 4 months. In fact, all of awhetter's issues are without reply for a long time. And Google's Chromium project is notorious for not accepting outside contributions.
So, I wouldn't call this "in-progress". The review link is valuable, though.

data/upstream.yaml
```diff
@@ -187,6 +199,10 @@ jabber-roster:
   repo: https://github.com/kparal/jabber-roster
   note: |
     Porting will start if deps are available.
+konversation:
+    status: in-progress
+    links:
+        homepage: https://konversation.kde.org/
```
encukou: Can you add a note or link that would explain? How can one check the progress?

Krejdom: https://quickgit.kde.org/?p=konversation.git&a=commit&h=285fc5f496363937b3676f33e7b468b8d85cce5e

torsava: I found this on the homepage:

> Konversation now installs an experimental Python scripting support package named 'konversation' into a subdirectory of its application data directory and appends all 'konversation/scripting_support/python' directories found in any KDE application data resource directories (i.e. within $KDEHOME, $KDEDIRS, etc.) to the PYTHONPATH environment variable available in the script execution context, thereby allowing Python scripts executed by Konversation to import the package. The package currently sports modules providing APIs for i18n support and D-Bus communication with Konversation.

> The 'media' script now requires Python 2.6 or higher and is compatible with Python 3.

The commit Dominika found relates to the 2nd paragraph—the 'media' script—therefore I think the package is only partially ported. Should I leave it as 'in-progress' or not?

encukou: I'd like to use the in-progress status as little as possible*, since for anybody who wants to help it's practically the same as idle. I think it would be much more useful to summarize the above into `notes` and keep it idle.

* This wasn't always the case, there's a lot of garbage in in-progress currently :(

torsava:
I'm sorry, I should have checked it more thoroughly. Thanks for the input!

encukou:
Thanks! Fixed YAML syntax and pushed as 6922407.

torsava:
What did you change so I can avoid it in the future? I can't see the diff.

data/upstream.yaml
```diff
@@ -187,6 +197,13 @@ jabber-roster:
   repo: https://github.com/kparal/jabber-roster
   note: |
     Porting will start if deps are available.
+konversation:
+    links:
+        homepage: https://konversation.kde.org/
+    note:
```
encukou: This needs to be `note: |` (note the pipe character) so the whole following block is treated as a string.

torsava: D'oh! Thanks!

PR #134, opened 2016-04-05
Removed 'status: released' tag of most of packages.

torsava:
Removed from packages which are:
a) already ported in Fedora, or
b) already tracked as mispackaged or in-progress by portingdb

It remains for 3 packages which cannot be moved to mispackaged due to unported
dependencies.


encukou:
Thank you!
Rebased and merged as cee8c74

PR #144, opened 2016-05-03
Note for pagure.

torsava:


encukou:
Merged as bf25c1a. Thanks!

PR #145, opened 2016-05-05
Added a script for finding 'idle' packages marked as 'released' upstream

torsava:
If you find a better name for the cli command (this would be the first one with an underscore) I'm open to suggestions.


torsava:
And please send me the output after you merge it so I can file these bugs.

encukou:
I put my suggestions in https://github.com/torsava/portingdb/pull/1 :)

PR #147, opened 2016-05-09
Dropped the package python-urlgrabber

torsava:
Based on information from Sheldon who tried to port the package.


encukou:
Thanks!
This unblocked `cas` and `livecd-tools`, could you check upstream status there?

torsava:
Neither looks ready for Python 3.

PR #151, opened 2016-05-19
Fixed reture value of the 'upstream-idle' cli script.

torsava:


encukou:
The doc says `Exits with error code 1 if such packages are found.`. That's on purpose – it's “bad” if such packages exist.

torsava:
Oh, I thought you said it's so you can use it in pipes and so on, which the current implementation prevents I believe.

encukou:
It shouldn't. What's the pipeline you're having trouble with?

torsava:
Never mind, I just had enabled zsh/bash option `pipefail`. Sorry!

PR #154, opened 2016-05-22
fix history graph when served over https

collinanderson:
example url: https://portingdb-encukou.rhcloud.com/history/


encukou:
Makes sense. Thanks!

PR #155, opened 2016-05-23
Adding a way to list packages tracked by PY3PATCH-REQUESTED

torsava:
So far I've only added a TrackingBug(s) class/table into the DB scheme.

### DIFF
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -24,14 +24,25 @@ from dnfpluginscore import _
 import bugzilla  # python-bugzilla
 
 BUGZILLA_URL = 'bugzilla.redhat.com'
+# Tracker bugs which are used to find all relevant package bugs
 TRACKER_BUG_IDS = [
     1285816,  # The Python 3 tracking bug
-    1322027   # The Python 3 Upstream Porting tracking bug
+    1322027,  # The Python 3 Upstream Porting tracking bug
 ]
-# Trackers of bugs that indicate the given package is mispackaged.
+# Trackers of bugs whose presence indicates the given package is mispackaged.
 MISPACKAGED_TRACKER_BUG_IDS = [
-    1285816   # The Python 3 tracking bug
+    1285816,  # The Python 3 tracking bug
 ]
+# Trackers of bugs whose presence gives us additional information about the
+# status quo.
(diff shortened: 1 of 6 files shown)


encukou:
Looks reasonable, but please add the change to the DNF plugin and the load code as well

torsava:
Of course!

dnf-plugins/py3query.py
```diff
@@ -24,13 +24,21 @@
...
 
 BUGZILLA_URL = 'bugzilla.redhat.com'
+# Tracker bugs which are used to find all relevant package bugs
 TRACKER_BUG_IDS = [
     1285816,  # The Python 3 tracking bug
-    1322027   # The Python 3 Upstream Porting tracking bug
+    1322027,  # The Python 3 Upstream Porting tracking bug
 ]
-# Trackers of bugs that indicate the given package is mispackaged.
+# Trackers of bugs wose presence indicates the given package is mispackaged.
```
encukou: `whose`

portingdb/tables.py
```diff
@@ -267,6 +267,27 @@ def __repr__(self):
                                        self.collection_package.name)
 
 
+class TrackingBug(TableBase):
+    u"""Tracking bugs associated with a package."""
+    __tablename__ = 'tracking_bugs'
+    __table_args__ = {'sqlite_autoincrement': True}
```
encukou: I don't think autoincrement is necessary. But please add a unique constraint.

portingdb/tables.py
```diff
@@ -267,6 +267,27 @@ def __repr__(self):
...
+
+    id = IDColumn()
+    collection_package_id = Column(
+        ForeignKey(CollectionPackage.id), nullable=False)
+    tracking_bug = Column(
+        Integer(), nullable=False,
+        doc='ID number of the tracking bug')
+
+    collection_package = relationship(
+        'CollectionPackage', backref=backref('tracking_bugs'))
```
encukou: Look at at the bottom of tables.py for a few examples of doing an N-to-M relationship with intermediate table.

torsava:
I think the pull request is ready.

encukou:
Looks good, thanks!

torsava:
My pleasure!

PR #156, opened 2016-05-24
Adding a page that will list Mispackaged packages ordered by last BZ comment

torsava:
Hi! This is a preview, it's not yet ready to be merged.

The data/DB scheme and loading of the data should be complete.

Now I just need to display it, but that'll have to wait till tomorrow.


encukou:
Looks good! two nitpicks:

* Could you name the column `last_update`? I think we can be specific here. (Using `note` for bug statuses was a bit of a mistake)
* In the JSON, could you use a RFC 3339 format (`YYYY-MM-DD HH:MM:SS`, generated with `str(datetime_value)`)? It's way more readable than a numeric timestamp.

torsava:
Are you in Amsterdam right now? :)

> Could you name the column last_update? I think we can be specific here. (Using note for bug statuses was a bit of a mistake)

Consider it done.

> In the JSON, could you use a RFC 3339 format (YYYY-MM-DD HH:MM:SS, generated with str(datetime_value))? It's way more readable than a numeric timestamp.

I was a bit stumped that DateTime isn't JSON serializable, so I chose the timestump which was easiest. I will switch it to the string, no problem.

torsava:
I rebased the commits, now they should adhere to your notes.

Take a look into `htmlreport.py`. I'm using a `@property` of Package called `last_link_update` to get the date of the last BZ comment. However, it is not possible to sort on properties in SQLAlchemy unless you write the property as a `@hybrid_property` in SQL expressions, which I'm not certain is possible in this case. Therefore I'm sorting it using Python, which is quite straight forward if possibly a bit slower.

As the page will be cached anyway I think this is ok, however, if you disagree I can bash my head against it some more. :)

portingdb/htmlreport.py
```diff
@@ -588,6 +589,82 @@ def ascending(p):
...
+    if requested:
+        # Filter only to packages where maintainer requested a patch
+        mispackaged = [pkg for pkg in mispackaged
+                if "https://bugzilla.redhat.com/show_bug.cgi?id=1333765"
+                in pkg.list_tracking_bugs]
+
+    # Sorting the packages based on their last_link_update
+    #   Note that sorting through the database is problematic, because
+    #   to sort on a @property you need to create a @hybrid_property which
+    #   needs to be written as sort of an SQL expression
```
encukou: I don't think that's true, you can sort by joining a table and sorting on it. `@hybrid_property` just makes some cases more convenient.
Anyway, it doesn't matter that much if sorting is done in SQL or Python.

dnf-plugins/py3query.py
```diff
@@ -268,7 +269,14 @@ def bug_namegetter(bug):
                 status = bug.status
                 if bug.resolution:
                     status += ' ' + bug.resolution
-                r.setdefault('links', {})['bug'] = [bug.weburl, status]
+                # Let's get the datetime of the last comment
+                last_comment_datetime = None
+                if bug.comments:
```
encukou: This makes a new request to Bugzilla for every bug. That's quite slow and inefficient.
Could you add `last_change_time` to `include_fields` above, and use `bug.last_change_time`? It's "last activity" rather than "last commit", but it's much cheaper to get.

portingdb/htmlreport.py
```diff
@@ -588,6 +589,82 @@ def ascending(p):
...
+
+    # Packages without a BZ report will be last
+    nonevalue = datetime.datetime.utcnow()
+    mispackaged = sorted(mispackaged,
```
encukou: mispackaged is a list, so you can call its `sort()` method rather than creating a new list.

portingdb/htmlreport.py
```diff
@@ -588,6 +589,82 @@ def ascending(p):
...
+        mispackaged = [pkg for pkg in mispackaged
+                if "https://bugzilla.redhat.com/show_bug.cgi?id=1333765"
+                in pkg.list_tracking_bugs]
+
+    # Sorting the packages based on their last_link_update
+    #   Note that sorting through the database is problematic, because
+    #   to sort on a @property you need to create a @hybrid_property which
+    #   needs to be written as sort of an SQL expression
+
+    # Packages without a BZ report will be last
```
encukou: Why is this? I think they should be hidden, or given a separate section. From the page it's not clear why there's suddenly no date for a bunch of packages.

portingdb/load.py
```diff
@@ -234,7 +239,7 @@ def load_from_directories(db, directories):
                 v['type'] = 'bug'
         print(set(v['type'] for v in values))
         bulk_load(db, values, tables.Link.__table__,
-                  key_columns=['collection_package_id', 'url'])
+                  key_columns=['collection_package_id', 'url', 'last_update'])
```
encukou: `last_update` shouldn't be here. The `collection_package_id`/`url` combination should be unique by itself.

portingdb/tables.py
```diff
@@ -139,6 +139,12 @@ def nonblocking(self):
     def list_tracking_bugs(self):
         return [tb.url for cp in self.collection_packages for tb in cp.tracking_bugs]
 
+    @property
+    def last_link_update(self):
+        values = [link.last_update for cp in self.collection_packages
+                for link in cp.links if link.last_update is not None]
+        return values[0] if values else None
```
encukou: Nitpick: I don't think this is one of the rare cases where the if-else operator is better than a full if statement.

portingdb/tables.py
```diff
@@ -308,6 +314,10 @@ class Link(TableBase):
     note = Column(
         Unicode(), nullable=True,
         doc='Type-specific note about the link')
+    last_update = Column(
+        DateTime(), nullable=True,
+        doc='Type-specific datetime field')
```
encukou: Please also update the doc :)

portingdb/templates/index.html
```diff
@@ -55,7 +55,11 @@ <h2 id="in-progress">Porting in Progress</h2>
             {% if mispackaged_packages %}
                 <h2 id="mispackaged">Mispackaged packages</h2>
                 <div>
-                    {{ len(mispackaged_packages) }} packages were found to have some kind of packaging problem:
+                    {{ len(mispackaged_packages) }} packages were found to have some kind of packaging problem.
+                </div>
+                <div>
+                    To find the best package to port, see <a href="{{ url_for('mispackaged') }}">packages ordered by last BugZilla comment</a>.<br/>
```
encukou: Why do you spell Bugzilla with a capital Z?

portingdb/templates/mispackaged.html
```diff
@@ -0,0 +1,56 @@
...
+    <div class="col-md-12">
+        <h1>Mispackaged packges</h1>
+        <p>
+            {% if not requested %}
+                You can see <i>all packages</i> or <a href="{{ url_for('mispackaged', requested=1) }}">packages where the maintainer requested a patch</a>.
+            {% else %}
+                You can see <a href="{{ url_for('mispackaged') }}">all packages</a> or <i>packages where the maintainer requested a patch</i>.
+            {% endif %}
+        </p>
+        <ul class="simple-pkg-list">
```
encukou: Could you use `<table>` for the table?

portingdb/htmlreport.py
```diff
@@ -588,6 +589,82 @@ def ascending(p):
...
+    # Render the page, pass the data
+    return render_template(
+        'mispackaged.html',
+        breadcrumbs=(
+            (url_for('hello'), 'Python 3 Porting Database'),
+            (url_for('mispackaged'), 'Mispackaged'),
+        ),
+        requested=bool(requested),
+        mispackaged=mispackaged,
+        ago=ago,
```
encukou: It would be a bit cleaner to use `ago` as a filter, see `app.jinja_env.filters` below.

portingdb/templates/mispackaged.html
```diff
@@ -0,0 +1,56 @@
...
+                    {{ pkglink(pkg) }}
+                    {% if not requested and
+                            "https://bugzilla.redhat.com/show_bug.cgi?id=1333765"
+                            in pkg.list_tracking_bugs %}
+                       (patch requested by maintainer)
+                    {% endif %}
+                    </span>
+
+                    {% if pkg.last_link_update %}
+                        {{ ago(pkg.last_link_update) }}
```
encukou: Could you use [`<time>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) here?
Also, a way to see the full datetime would be helpful. Maybe a tooltip with [`title`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title)?

portingdb/htmlreport.py
torsava: My thinking is thus: There really shouldn't be any bugs without a BZ report, so list them last so people are least likely to go work on them for the short while while they are bugless. But I don't like hiding them, I don't think it's necessary.

I plan to make a script (like upstream-idle) which automatically finds all such bugs and reports them to us for bug filing during the update process.

dnf-plugins/py3query.py
torsava: That would mean that if someone changes some little detail—for example if we change the Blocks field it—looks like it's active. I'll try to measure how much slower this makes the execution, because I don't remember it being that significant, but maybe it was.

portingdb/htmlreport.py
torsava: Well, it's a bit problematic, because some packages have more Link-bug fields, some of which are Null on the last_update and some of which are not, and some packages just have a Link-bug with a Null on last_update. I'll remove the comment if it doesn't matter if I sort it in Python anyway.

portingdb/htmlreport.py
torsava: If not requested: then it's not yet a list, though I could make it so in an else branch above to avoid making two lists. I'll do that.

portingdb/htmlreport.py
torsava: Ok, I'll look into those!

portingdb/load.py
torsava: Roger.

portingdb/tables.py
torsava: Well, I don't believe there is a rarity of such cases, but I'll conform to your opinion as it is your code base. :)

portingdb/tables.py
torsava: Noted.

portingdb/templates/index.html
torsava: Huh, now you stumped me. I think it has to do with my love for CamelCase that somehow made me believe it was the official spelling of the name. :-o

It shall be rectified!

portingdb/templates/mispackaged.html
torsava: Hm, you can have a <li> elements inside a <table>? I'll investigate.

portingdb/templates/mispackaged.html
torsava: I'll make it so!

portingdb/htmlreport.py
```diff
@@ -588,6 +589,82 @@ def ascending(p):
...
+    """Displays roughly how long ago the date was in a human readable format"""
+    now = datetime.datetime.utcnow()
+    diff = now - date
+
+    # Years
+    if diff.days >= 365:
+        if diff.days >= 2 * 365:
+            return "{} years ago".format(math.floor(diff.days / 365))
+        else:
+            return "a year ago"
```
hroncok: 365+364 days is still a year ago?

torsava:
Thanks for all the notes, I shall work them in on Monday!

portingdb/templates/mispackaged.html
hroncok: Yes, you can.

portingdb/htmlreport.py
torsava: Technically yeah. I didn't work in rounding for the years as you really don't need a big precision there. The point of the list is ordering by time, the specific values are just for quick orientation. I'm open to PRs if you think it's important :)

portingdb/templates/mispackaged.html
torsava: Hm, I don't know why but I was pretty sure you couldn't. Probably IE5.5 for which I started writing HTML didn't display it correctly and thus my memory cemented this as impossible. :)

hroncok: http://www.w3schools.com/tags/tryit.asp?filename=tryhtml_table_elements

torsava: Well of course you can do that. But we need to put each `<li>` inside its own `<td>`, that means the `<ul>` will have to encompass the whole `<table>`. That I was instinctively sure cannot be done, but again, I learned HTML against the IE5.5 which was the most horrible browser on the planet. :)

dnf-plugins/py3query.py
encukou: Yes, `last_change_time` might be even better than last commit. It's definitely not worse, IMO.

portingdb/htmlreport.py
encukou: OK, that works.

portingdb/templates/mispackaged.html
encukou: Don't use `<li>` at all. See how the in-progress table on the main page, or the dependency table on package pages, are done.

torsava: Oh right, of course! Sorry, my brain wasn't fully turned on on Saturday morning.

dnf-plugins/py3query.py
torsava: Ok, I will put that in.

torsava:
Hopefully this addresses all the issues!

torsava:
@encukou When you feel it's ready, please just leave a note, I'd like to do the merge and deployment.

portingdb/templates/mispackaged.html
```diff
@@ -0,0 +1,70 @@
...
+{%- endblock titlecontent %}
+
+{% block style %}
+{{ super() }}
+
+body {
+  /*font: 10px sans-serif;*/
+}
+
+{% endblock style %}
```
encukou: This block is no longer needed.

encukou:
Looks good, just remove `{% block style %}` in `mispackaged.html`.
Once that's fixed, please merge with a merge commit (`--no-ff`), so it's apparent that all the commits are related.

portingdb/templates/mispackaged.html
torsava: It looks like I can safely remove all that except the first line: `{% extends "_base.html" %}`.

If I remove that, I get an error: `jinja2.exceptions.UndefinedError: 'pkglink' is undefined`, because the `pkglink` macro is defined there. Thus I think the first line should stay on.

torsava:
Merged, thanks for the review and guidance!

I'll push it to production with tomorrow's update.

PR #157, opened 2016-05-31
New cli.py script: bugless-mispackaged

torsava:
This will find mispackaged packages that are without a Bugzilla report.

It should be used after portingdb update each day like `upstream-idle`.


encukou:
Looks good, please merge!

torsava:
Merged, thanks!

PR #159, opened 2016-06-13
Package page: Added last_update and PY3PATCH-REQUESTED indicators

torsava:
There might be a better place to put the REQUESTED text, so if you think of something, let me know please.


encukou:
I think this is a good place. Just please use `<time>` like on the other page.

torsava:
Ah, damn my wretched memory, of course! I amended the commit accordingly.

PR #160, opened 2016-06-15
Improvements, mainly to the howto

encukou:
@torsava, do these look OK to you?

### DIFF
--- a/data/config.yaml
+++ b/data/config.yaml
@@ -17,14 +17,14 @@ name: Fedora
 # The string '$RANDOMPKG' will be replaced by a random "idle" package, if
 # it occurs in the text.
 main-instructions: |
-    <div>
+    <p>
         This is a dashboard to track progress of porting
         Fedora packages to Python 3.
-    </div>
-    <div style="padding-top: 7pt;">
+    </p>
+    <p>
         Read me: <a href="$HOWTO_LINK">So you want to contribute?</a><br/>
-    </div>
-    <div style="padding-top: 7pt;">
+    </p>
+    <p>
         If you would like to jump in and help port, do this:
         <ul>
             <li>
@@ -42,20 +42,20 @@ main-instructions: |
                 <a href="https://github.com/fedora-python/portingdb/issues">file an issue</a>.
             </li>
         </ul>
-    </div>
+    </p>
 
 # A note about "groups" and their role in the report can be put in
 # "groups-note", as HTML.
 groups-note: |
-    <div>
+    <p>
         If you're interested in porting a bigger, self-contained piece
         of Fedora, choose one of these.
-    </div>
-    <div>
+    </p>
+    <p>
         Note that some of these are still incomplete early examples.
         Please propose changes at
         <a href="https://github.com/fedora-python/portingdb/blob/master/data/groups.yaml">Github</a>.
-    </div>
+    </p>
 
 # Notes on how the data is updated and extra information can be put in
 # "extra-sidebar", in HTML. It'll appear on the main mage, in the sidebar.
--- a/data/config.yaml
+++ b/data/config.yaml
@@ -24,7 +24,7 @@ main-instructions: |
     <p>
         Read me: <a href="$HOWTO_LINK">So you want to contribute?</a><br/>
     </p>
-    <p>
+    <div>
         If you would like to jump in and help port, do this:
         <ul>
             <li>
@@ -42,7 +42,7 @@ main-instructions: |
                 <a href="https://github.com/fedora-python/portingdb/issues">file an issue</a>.
             </li>
         </ul>
-    </p>
+    </div>
 
 # A note about "groups" and their role in the report can be put in
 # "groups-note", as HTML.
--- a/portingdb/htmlreport.py
+++ b/portingdb/htmlreport.py
@@ -46,38 +46,35 @@ def hello():
             'data': data,
         }
 
-    total_pkg_count = queries.packages(db).count()
-
     # Main package query
 
     query = queries.packages(db)
+    total_pkg_count = query.count()
 
-    active, query = queries.split(query, tables.Package.status == 'in-progress')
+    active = query.filter(tables.Package.status == 'in-progress')
     active = queries.order_by_weight(db, active)
     active = queries.order_by_name(db, active)
     active = active.options(subqueryload('collection_packages'))
     active = active.options(subqueryload('collection_packages.links'))
 
-    done, query = queries.split(query, tables.Package.status == 'released')
+    done = query.filter(tables.Package.status == 'released')
     done = queries.order_by_name(db, done)
 
-    dropped, query = queries.split(query, tables.Package.status == 'dropped')
+    dropped = query.filter(tables.Package.status == 'dropped')
     dropped = queries.order_by_name(db, dropped)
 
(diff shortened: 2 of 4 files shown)


data/config.yaml
```diff
@@ -17,14 +17,14 @@ name: Fedora
...
-    </div>
-    <div style="padding-top: 7pt;">
+    </p>
+    <p>
         Read me: <a href="$HOWTO_LINK">So you want to contribute?</a><br/>
-    </div>
-    <div style="padding-top: 7pt;">
+    </p>
+    <p>
         If you would like to jump in and help port, do this:
```
torsava: The change creates a vertical space between this line and the list below. That means we have 4 visual blocks of text in this section instead of 3, which makes it look a bit too visually complex. That in turn might make some people skip reading it. Though that's only my speculation and how I like to format things.

portingdb/htmlreport.py
```diff
@@ -46,39 +46,36 @@ def hello():
...
+    blocked = query.filter(tables.Package.status == 'blocked')
     blocked = blocked.options(subqueryload('requirements'))
     blocked = queries.order_by_name(db, blocked)
 
-    ready, query = queries.split(query, tables.Package.status == 'idle')
+    ready = query.filter(tables.Package.status == 'idle')
     ready = ready.options(subqueryload('requirers'))
     ready = queries.order_by_name(db, ready)
 
-    assert query.count() == 0
```
torsava: That was pretty clever, I haven't noticed it worked like that :)

portingdb/htmlreport.py
```diff
@@ -479,34 +480,29 @@ def howto():
     db = current_app.config['DB']()
     query = queries.packages(db)
 
-    # Count the idle and blocked packages
-    idle_query, query = queries.split(query, tables.Package.status == 'idle')
-    idle_len = idle_query.count()
-    blocked_query, query = queries.split(query, tables.Package.status == 'blocked')
+    # Count the blocked packages
+    blocked_query = query.filter(tables.Package.status == 'blocked')
     blocked_len = blocked_query.count()
```
torsava: Is this better then just doing `len(blocked)`?

encukou: If there was a `blocked`, then it wouldn't be better. Here the count method it generates a `SELECT COUNT()` query without getting the actual list of objects.

portingdb/htmlreport.py
```diff
@@ -479,34 +480,29 @@ def howto():
...
-    # Upstream status: Released
+    # Status objects
     query = db.query(tables.Status)
-    query = query.filter(tables.Status.ident == 'released')
-    query = query.join(tables.CollectionPackage)
-    query = query.filter(
-            tables.CollectionPackage.collection_ident == 'upstream')
-    released_status = query.first()
+    mispackaged_status = query.get('mispackaged')
+    released_status = query.get('released')
```
torsava: Aren't there multiple 'released' statuses for different collections? Do you count on them being the same?

encukou: The status object (i.e. row in database) is the same.

encukou: i.e. the `tables.Status` is the same, there are `tables.CollectionStatus` but those just add a collection-dependent description.

torsava: Oh I see! Could they be theoretically different or have I misread the scheme?

encukou: `table.sStatus.ident` is the table's only primary_key, so they can't ever be two different statuses with the same identifier :)

portingdb/templates/howto.html
```diff
@@ -33,13 +33,15 @@ <h2 id="package-it">Package it</h2>
...
         </p>
         <h3>Make a Patch</h3>
         <p>
-            In either case, look a bit further down still and find a link to the associated Bugzilla report (under <i>Links→bug</i> in the sidebar). Go there and join the discussion. Find out what's the status quo and if anyone is working on it. If not, be the one that submits a patch!
+            In either case, look a bit further down on the package page in PortingDB and find a link to the associated Bugzilla report (under <i>Links→bug</i> in the sidebar). Go there and join the discussion. Find out what's the status quo and if anyone is working on it. If not, be the one that submits a patch!
+        </p><p>
+            If you don't find a link to Bugzilla, you'll need to investigate the issue and <a href="https://fedoraproject.org/wiki/User:Pviktori/Python_3_Bug_Filing">file a bug</a> yourself. A hint will usually provided just under the Fedora heading on the package page. When filing a bug, please make it block the <b>PYTHON3</b> tracking bug.
```
torsava: Good addition. Though I would omit the:

> When filing a bug, please make it block the PYTHON3 tracking bug.

That is best left for the file-a-bug wiki page, which I am planning to give a bit of a polish too.

portingdb/templates/howto.html
```diff
@@ -61,7 +63,7 @@ <h2 id="investigate-annotate">Investigate & Annotate</h2>
         </p>
         <h4>Blocked packages</h4>
         <p>
-            The <a href="{{ url_for('hello') }}#blocked">{{ blocked_len }} blocked packages</a> are a bit similar, but not quite. These are Python 2–only packages that, however, cannot yet be ported because some of their dependencies are still not ported to Python 3 in Fedora.
+            The <a href="{{ url_for('hello') }}#blocked">{{ blocked_len }} blocked packages</a> are a bit similar, but not quite. These are Python 2–only packages that, cannot yet be ported because some of their dependencies are still not ported to Python 3 in Fedora.
```
torsava: I agree, it sounds better. However, the comma became superfluous.

> that, cannot

torsava:
Big thanks for the polish. The Bootstrap styling looks much better, I should give it a whirl sometimes.

So besides the few notes, looks good to me!

data/config.yaml
encukou: OK, I changed the last `<p>` back to `<div>`

portingdb/templates/howto.html
encukou: OK, I left that out, and reworded it a bit more.

portingdb/templates/howto.html
```diff
@@ -41,7 +41,7 @@ <h2 id="package-it">Package it</h2>
         <p>
             In either case, look a bit further down on the package page in PortingDB and find a link to the associated Bugzilla report (under <i>Links→bug</i> in the sidebar). Go there and join the discussion. Find out what's the status quo and if anyone is working on it. If not, be the one that submits a patch!
         </p><p>
-            If you don't find a link to Bugzilla, you'll need to investigate the issue and <a href="https://fedoraproject.org/wiki/User:Pviktori/Python_3_Bug_Filing">file a bug</a> yourself. A hint will usually provided just under the Fedora heading on the package page. When filing a bug, please make it block the <b>PYTHON3</b> tracking bug.
+            If you don't find a link to Bugzilla, you'll need to investigate the issue and <a href="https://fedoraproject.org/wiki/User:Pviktori/Python_3_Bug_Filing">file a bug</a> yourself. A hint on the nature of the problem will usually provided just under the Fedora heading on the package page.
```
torsava: > A hint on the nature of the problem will usually provided

I just noticed there's a missing 'be' there.

encukou: Thanks! I force-pushed a fix :)

torsava:
Perfect!

PR #162, opened 2016-06-27
Added missing packages into the OpenStack group

ari3s:
Added missing packages into the OpenStack group according to https://wiki.openstack.org/wiki/Python3


encukou:
Thank you!

PR #165, opened 2016-07-11
python-tuskarclient has been deprecated upstream

Krejdom:
python-tuskarclient is currently in the "green section", it should be dropped.
https://bugzilla.redhat.com/show_bug.cgi?id=1309311


encukou:
The package was removed from Fedora; it'll be unlisted from portingdb automatically soon (after the relevant rawhide compose gets to mirrors).

PR #168, opened 2016-07-15
trytond-account-invoice-history  py3 upstream available

ari3s:


encukou:
Thanks!
Added `-history` to the package name and pushed as aa53e518303afdfe561ba6aff1c5662885465bbf

PR #169, opened 2016-07-15
Fix status of M2Crypto.

mcepl:
M2Crypto is not completely dead, and yes it is not the only Python bindings for OpenSSL, but it still has surprisingly large user base.

### DIFF
--- a/data/fedora-update.yaml
+++ b/data/fedora-update.yaml
@@ -211,13 +211,15 @@ notify-python:
   note: |
     Last update from upstream was in 2006.
 m2crypto:
-    status: dropped
     note: |
       Suggested replacement: `python-cryptography` (or Python standard library).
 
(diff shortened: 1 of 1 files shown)


encukou:
So, what's the state of porting m2crypto? Will it be ready any time soon? Is there a link to some discussion/pull request/roadmap that users can check to get updates on the porting effort?

mcepl:
merge request is https://gitlab.com/m2crypto/m2crypto/merge_requests/65, currently the test suite still doesn't pass completely (https://gitlab.com/m2crypto/m2crypto/builds/2291813 and some tests are skipped). We are adding PEP-484 typehints to all functions of M2Crypto, but finishing of this python3 merge request is next on my TODO list.

encukou:
I added the link in ad571840d01f7e3af215897b3d970b9a12391d46.

PR #174, opened 2016-07-31
Not maintained packages

ari3s:
Not maintained packages:

- python-django-authenticator
    Repository: https://github.com/ASKBOT/django-authenticator
    Not ready for use in production, not maintained since 2011.

- python-django-dpaste
    Repository: https://github.com/digitaljhelms/django-paste
    This package is no longer in development. Please see the successor dpaste.

- python-django-staticfiles
     The package is now part of Django (since 1.3) as django.contrib.staticfiles.

### DIFF
--- a/data/fedora-update.yaml
+++ b/data/fedora-update.yaml
@@ -418,6 +418,11 @@ python-dateutil15:
     note: |
         This is an old version of the dateutil library.
         Replacement: `python-dateutil`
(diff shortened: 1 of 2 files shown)


encukou:
Thanks!
I've marked `python-django-staticfiles` as dropped – it has a replacement in Fedora :)

PR #175, opened 2016-08-02
Upstream bugs

ari3s:
Upstream bugs related to

- python-django-openid-auth (there is not considered that python3-openid is available for Python 3)

- python-restkit

### DIFF
--- a/data/upstream.yaml
+++ b/data/upstream.yaml
@@ -395,6 +395,11 @@ python-django-haystack:
   links:
     homepage: http://haystacksearch.org/
     repo: https://pypi.python.org/pypi/django-haystack/2.5.0
+python-django-openid-auth:
+  links:
+    homepage: https://launchpad.net/django-openid-auth
+    repo: https://pypi.python.org/pypi/django-openid-auth
+    bug: https://caniusepython3.com/project/django-openid-auth
 python-django-pipeline:
   status: released
   links:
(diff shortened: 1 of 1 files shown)


data/upstream.yaml
```diff
@@ -395,6 +395,11 @@ python-django-haystack:
   links:
     homepage: http://haystacksearch.org/
     repo: https://pypi.python.org/pypi/django-haystack/2.5.0
+python-django-openid-auth:
+  links:
+    homepage: https://launchpad.net/django-openid-auth
+    repo: https://pypi.python.org/pypi/django-openid-auth
+    bug: https://caniusepython3.com/project/django-openid-auth
```
encukou: Sorry for the delay (blame summer!)

Where does the caniusepython3.com link come from? As far as I can see it's not linked from the project.

ari3s:
You can see more information at [1].

As I checked the last available version of python-django-openid-auth at the repo [2], there is the blocker [3] presented in the source code.

[1] https://pypi.python.org/pypi/caniusepython3
[2] https://pypi.python.org/pypi/python-openid/2.2.5
[3] https://caniusepython3.com/project/django-openid-auth

encukou:
But python-openid is ported, in a fork [1] [2].
Where does the django-openid-auth source link to caniusepython3? It should be fixed there.

I submitted the python-openid fork to caniusepython3 [3], so on their next update it should no longer be marked as  blocked there.

[1] http://fedora.portingdb.xyz/pkg/python-openid/
[2] https://github.com/necaris/python3-openid
[3] https://github.com/brettcannon/caniusepython3/pull/153/files

ari3s:
As I know there is no direct link from django-openid-auth project, so the bug can be ignored.

encukou:
OK, thanks for clearing that up, and sorry for the misunderstanding.

PR #176, opened 2016-08-02
Dead upstream packages

ari3s:


data/fedora-update.yaml
```diff
@@ -449,18 +455,73 @@ python-django-dpaste:
       Project is retired, replaced by not-yet-packaged `dpaste`
     links:
       homepage: https://pypi.python.org/pypi/django-dpaste/0.2.4
+python-django-dynamite:
+    dead-upstream: true
+    links:
+      homepage: https://github.com/z4r/django-dynamite
+    note: |
+      Not maintained since 2009
```
encukou: Where does `2009` come from? There was a "0.4" release in 2011.

ari3s: You are right, it is a typo. The latest activity was there in 2012. Thanks for your check. Please could you make the small correction in the note directly?

encukou: Sure. The problem is that the year seems off on lots of packages in this PR. So I'm curious how you decide on the years.

ari3s:
I used last commits at referenced homepages, or in the case of python-batchhttp where such a commit was not relevant to its newest available version (on PyPI) then it is this version year.

However, there is a additional typo at python-django-typepadapp where the year should be 2010.

encukou:
I've changed the wording to "last upstream commit" or "last release" to make it clearer. I'd like to use "not maintained" and similar wordings in cases where the author explicitly says it's abandoned.

Thanks for gathering the information!

PR #184, opened 2016-08-25
Update status on fabric

rodrigc:
I submitted https://bugzilla.redhat.com/show_bug.cgi?id=1370016 to request that Fedora use @mathiasertl 's [Python 3 compatible fork of fabric](https://github.com/mathiasertl/fabric).

### DIFF
--- a/data/fedora.json
+++ b/data/fedora.json
@@ -5662,9 +5662,9 @@
     ],
     "links": {
       "bug": [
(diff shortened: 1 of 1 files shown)


encukou:
Hi! Thanks for filing the issue!
I added `PYTHON3` as a blocker in Bugzilla, so this was picked up by the update script.
If you need to change anything more, please do it in `fedora-update.yaml`. The json file is generated automatically.

PR #185, opened 2016-09-03
python-landslide's active fork available

ari3s:


encukou:
Sorry, this is not in scope for portingdb.
The original python-landslide is already py3-compatible upstream. Darkslide is not a fork that adds Python 3 compatibility, but one that fixes bugs and adds  – and even removes! – features.

Anyone can package Darkslide for Fedora, and perhaps the landslide maintainer would want to orphan landslide after that. But portingdb is not the place to suggest that people should switch to Darkslide.

PR #194, opened 2016-11-08
Two additional packages py3 compatible

ari3s:
- `python-openstack-nose-plugin`: it is works in Ubuntu [[http://packages.ubuntu.com/xenial/python3-openstack.nose-plugin](url)]
- `python-webpy`

### DIFF
--- a/data/upstream.yaml
+++ b/data/upstream.yaml
@@ -577,6 +577,10 @@ python-oslo-vmware:
   status: released
   links:
     homepage: https://pypi.python.org/pypi/oslo.vmware
+python-openstack-nose-plugin:
+  status: released
+  links:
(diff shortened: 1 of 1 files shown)


encukou:
Thanks Honza!
I see these are `idle`, not `blocked`, so we should file Bugzillas for these rather than just marking them in portingdb.

@irushchyshyn, can you file the Bugzillas (or tag existing ones, if there are any)?

irushchyshyn:
@encukou Ok, will do.

irushchyshyn:
Created Bugzillas:
https://bugzilla.redhat.com/show_bug.cgi?id=1393781
https://bugzilla.redhat.com/show_bug.cgi?id=1393840

PR #198, opened 2016-11-11
freeipa is upstream Python 3 available

Krejdom:


encukou:
Not really. There's some support, but it's broken (untested).
FreeIPA is a bit of a special snowflake :)

PR #199, opened 2016-11-11
Added python-tgext-admin and python-tgext-crud

ari3s:
`python-tgext-admin`: [https://pypi.python.org/pypi/tgext.admin/](https://pypi.python.org/pypi/tgext.admin/)
`python-tgext-crud`: [https://pypi.python.org/pypi/tgext.crud/](https://pypi.python.org/pypi/tgext.crud/)


encukou:
Thank you, Honza!

Fixed YAML syntax and merged as #974c425.

PR #201, opened 2016-11-15
Update Fedora data

irushchyshyn:
**_Changes:_**
idle -> mispackaged (3)
    `python-sprox`
    `python-tgext-admin`
    `python-tgext-crud`

Created Bugzilla tickets for packages in 974c425c15a72e304be0441daa076a4a4a7874a5 and `python-sprox` which is their dependency and supports Python 3 as well.


encukou:
We usually don't open bugs for packages that don't have all dependencies ported *in Fedora* already. I guess it's not a big deal here, since they all have the same maintainer.

Could you mention in the two bugs that `python-sprox` needs to be updated first, and make the `python-sprox` bug block them?

irushchyshyn:
@encukou Thanks for letting me know. Updated the bugs.

PR #207, opened 2016-11-29
livecd-tools has been ported to DNF and Python 3

Conan-Kudo:
The pull request is under review.


encukou:
Wow, that looks like a solid amount of work. Thanks for helping the port!

Conan-Kudo:
@encukou I'm glad to! :)

PR #213, opened 2016-12-15
Remove the in-progress  status in favor of more visible Bugzilla last-change date

encukou:
The `in-progress` status isn't really useful, since it mostly tracks when people signed up to do some work – which doesn't necessarily mean they'll actually do the work in finite time.

Stop generating `in-progress`, hide it (along with any other unused statuses) from the main page, and instead show the table from http://portingdb-encukou.rhcloud.com/mispackaged/ right on the main page.

### DIFF
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -288,10 +288,8 @@ class Py3QueryCommand(dnf.cli.Command):
                         r.setdefault('tracking_bugs', []) \
                                 .append(BUGZILLA_BUG_URL.format(tb))
 
-                if any(tb in bug.blocks for tb in MISPACKAGED_TRACKER_BUG_IDS):
-                    if r.get('status') == 'idle' and bug.status != 'NEW':
-                        r['status'] = 'in-progress'
-                    elif r.get('status') == 'idle' and bug.status == 'NEW':
+                if (any(tb in bug.blocks for tb in MISPACKAGED_TRACKER_BUG_IDS)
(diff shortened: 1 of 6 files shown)


encukou:
@irushchyshyn, does this look OK?

dnf-plugins/py3query.py
```diff
@@ -289,9 +289,7 @@ def key(bug):
                                 .append(BUGZILLA_BUG_URL.format(tb))
 
                 if any(tb in bug.blocks for tb in MISPACKAGED_TRACKER_BUG_IDS):
-                    if r.get('status') == 'idle' and bug.status != 'NEW':
-                        r['status'] = 'in-progress'
-                    elif r.get('status') == 'idle' and bug.status == 'NEW':
+                    if r.get('status') == 'idle' and bug.status == 'NEW':
```
irushchyshyn: We probably no longer need to check if `bug.status == 'NEW'`, because all the packages that were displayed as `in-progress` will remain `idle`.

encukou:
I'd like to avoid removing/breaking URLs in existing apps.
In this case it's probably irrelevant since it looks like we lost the domain name, but still, removing without replacement is bad style.

I've made all internal links point to the section of the front page, though.

irushchyshyn:
Thanks for the changes and the explanation, makes sense.
Looks good to me 👍 
Am I entitled to merge?

irushchyshyn:
Merged!

PR #222, opened 2017-01-06
python-django-stopforumspam is upstream Python 3 available

ari3s:
Bugzilla can be open for python-django-stopforumspam, version 1.7, to provide a Python 3 package for Fedora.


irushchyshyn:
Thanks @ari3s!
Opened Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1410815

encukou:
The change will appear on portingdb.xyz in the next update.

Issue #132, opened 2016-04-01
Track packages with py2 only binaries


hroncok:
We could automatically find packages that provide 2 subpackages (for 2 and 3) but only the python2 subpackage has something in /usr/bin.

Than we could mark them as mispackaged.

I think we have plenty of those packages in Fedora. See for example https://bugzilla.redhat.com/show_bug.cgi?id=1323256

encukou:
The DNF plugin is currently pretty naïve when it comes to finding corresponding py2/py3 packages: basically, it just compares the number of subpackages with py2 dependencies to the number of subpackages with py3 dependencies. (This gives false positives, e.g. [authconfig](http://fedora.portingdb.xyz/pkg/authconfig/) is now categorized manually.)
This would need some more detailed analysis. But getting the info about file provides shouldn't be hard.

hroncok:
I'm not very familiar with portingdb internals, but the idea with repoquery is:

    repoquery -l python-pygments --releasever rawhide | grep ^/usr/bin
    repoquery -l python3-pygments --releasever rawhide | grep ^/usr/bin

Where the first provides results and the second one does not. And we already know what py2 and py3 subpackages are generated form the SRPM.

hroncok:
We can be naive as well and countfiles in /usr/bin for both py2 and py3 packages and compare that somehow. We can raise the flag when py3 packages have none of such files, but py2 packages do.

irushchyshyn:
Created a PR #249 with the implementation for this.

As a result, there are around 30 packages affected. The list of packages:

- PyXB
- ahkab
- gdal
- lcgdm
- netcdf4-python
- pystache
- python-acme
- python-aexpect
- python-bugzilla
- python-docutils
- python-doit
- python-fiat
- python-gearbox
- python-igor
- python-igraph
- python-jsonschema
- python-ly
- python-ndg_httpsclient
- python-ngram
- python-oslo-policy
- python-oslo-privsep
- python-oslo-rootwrap
- python-pbr
- python-py2neo
- python-pygments
- python-rosdep
- python-rosinstall
- python-whisper

hroncok:
I've recently came across `python-docutils` (`rst2html`) and wanted to report. Will you do some mass bug filling?

irushchyshyn:
I was thinking, if we want to do mass bug filling if this gets merged, or if we should first send some warning to the mailing list?

hroncok:
[...]

PR #249, opened 2017-02-28
Identify packages with py2 only binaries

irushchyshyn:
Automatically find packages where only the python2 subpackage has something in `/usr/bin`.

This is a naive approach, which simply counts the number of files in `/usr/bin` for both py2 and py3 packages and marks as mispackaged those, where only the py2 subpackage contains binaries (issue #132).

### DIFF
--- a/data/fedora-update.yaml
+++ b/data/fedora-update.yaml
@@ -99,6 +99,12 @@ cwiid:
 dpdk:
     note: |
         Python 3 support committed but not yet released. (March 2016)
+dreampie:
+    status: idle
+    note: |
+        There's a python2 and a python3 subpackage, but they do different
+        things. dreampie-python3 is a plugin to run python3 interpreter
+        from within dreampie.
 drobo-utils:
     dead-upstream: true
 dtrx:
@@ -541,6 +547,11 @@ python-halite:
         "Halite is officially retired."
 python-HTMLgen:
     dead-upstream: true
+python-igor:
+    note: |
+        The package is functional under python3, but the executable output
+        is  slightly nicer under python2. So it is on py2 on purpose
+        according to the maintainer.
 python-ipaddr:
     status: dropped
     note: |
@@ -820,6 +831,9 @@ python-webtest1.3:
       ported to the current version of the library as part of porting to
       python3.
       Replacement: `python-webtest`
+python-whisper:
+    note: |
+        Whisper as library is python3 compatible, but the executables not yet.
 python-zbase32: &pyutil-zbase32-cycle
     nonblocking: true
     note: |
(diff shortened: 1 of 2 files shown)


dnf-plugins/py3query.py
```diff
@@ -129,25 +129,45 @@ def printer(i, name):
     print(file=sys.stderr)
 
 
+def binaries(packages):
+    """Return the total number of binaries in the packages."""
+    binaries = [
+        filepath for pkg in packages
+        for filepath in pkg.files
+        if filepath.startswith('/usr/bin')]
```
hroncok: or '/usr/sbin'

irushchyshyn: done in  	05a654c

encukou:
I'm swamped by other things but should get to the review this/next week. Let me know if this blocks something, I'll raise the priority.

irushchyshyn:
Thanks @encukou , this/next week is fine. This does not block anything.

encukou:
This looks good technically.
Before it's merged, could you go through the [mass filing procedure](https://fedoraproject.org/wiki/Mass_bug_filing)?, and add the resulting tracker bug to ADDITIONAL_TRACKER_BUGS?
Let me know if you need anything (e.g. a review of the mail's text).

irushchyshyn:
I have created tickets for the packages. The tracker bug is available here: https://bugzilla.redhat.com/show_bug.cgi?id=1432186

The tickets were created for:

- PyXB
- ahkab
- gdal
- lcgdm
- pystache
- python-aexpect
- python-bugzilla
- python-docutils
- python-doit
- python-fiat
- python-gearbox
- python-igraph
- python-jsonschema
- python-ly
- python-ngram
- python-oslo-policy
- python-oslo-privsep
- python-oslo-rootwrap
- python-pbr
- python-py2neo
- python-rosdep
- python-rosinstall

Notes added for the following packages, as the packagers replied that Py2 executables are preferred:

- python-whisper
- python-igor

I suggest to undo marking those packages as mispackaged, as the policy is so far only encouraged. We can change it when it is forced.

encukou:
Thanks!

data/fedora-update.yaml
```diff
@@ -541,6 +547,11 @@ python-halite:
         "Halite is officially retired."
 python-HTMLgen:
     dead-upstream: true
+python-igor:
+    note: |
+        The package is functional under python3, but the executable output
+        is  slightly nicer under python2. So it is on py2 on purpose
```
QuLogic: What does "nicer" mean? How is the output different on different Python versions?

irushchyshyn: This was the reply of the maintainer of the package in the mailing list: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/P3WPHS57OHH26XBLWIGFY4BJD35RU2VO/

PR #263, opened 2017-03-24
graphviz is Python 3 upstream available

ari3s:


data/upstream.yaml
```diff
@@ -215,6 +215,11 @@ gobject-introspection:
   links:
     homepage: https://wiki.gnome.org/Initiatives/GnomeGoals/Python3Porting
     bug: https://bugzilla.gnome.org/show_bug.cgi?id=679438
+graphviz:
+  status: released
+  links:
+    homepage: https://pypi.python.org/pypi/graphviz
+    repo: https://github.com/xflr6/graphviz
```
irushchyshyn: I believe that the repository for [graphviz](http://fedora.portingdb.xyz/pkg/graphviz/) is https://github.com/ellson/graphviz/.
The version in the spec file is [2.40.0](http://pkgs.fedoraproject.org/cgit/rpms/graphviz.git/tree/graphviz.spec#n51),  and the current version in https://github.com/xflr6/graphviz is 0.6. Those two seem to be different packages.

ari3s: You are right, the same name made me a confusion. I revert the commit, and make homepage/repo appropriate update to avoid the confusion for future.

irushchyshyn: Thanks a lot @ari3s !

Issue #268, opened 2017-03-30
Track Python 2 mispackaged packages


irushchyshyn:
Create a separate status for Python 2 (sub)packages which are named `python-foo` instead of `python2-foo` (with provides) and should be renamed in the nearest future.

There are packages like this across all statuses, so suggestions on displaying those in portingdb can be added to comments. 

Update:

* Name status should be a boolean field
* There should be 2 statuses: Correctly named and Incorrectly named
* Collect also information on cases like `sympy` (`foo`, `python3-foo`), which should also go into incorrectly named group

encukou:
Let's have the data before thinking about how to display it. Visualizations always depend on the data.

irushchyshyn:
TODO:
Prioritize misnamed packages, depending on how many other packages list them in `Requires`.

irushchyshyn:
Closing the ticket, all implemented in the linked Pull Requests.
Prioritization is no longer needed, as most misnamed packages have already been renamed. Please reopen if needed.

PR #273, opened 2017-04-04
Collect data for incorrectly named Python 2 packages

irushchyshyn:
This PR is **step 1** for #268 .

- [x] Create a separate status to track Python 2 renaming progress.
- [x] Store it in database as `py2status` column for `Package` and `CollectionPackage`
- [x] Collect it via dnf plugin and save to`fedora.json` to keep track of changes

Note: the database scheme is changed, so the db needs to be reloaded.
Result: There are 897 packages which are to be renamed to `python2-` .

### DIFF
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -174,6 +174,17 @@ def set_status(result, pkgs, python_versions):
         else:
             result['status'] = 'idle'
 
+    set_py2status(result, pkg_by_version)
+
+
+def set_py2status(result, pkg_by_version):
+    """Check if Python 2 subpackages are correctly named."""
+    result['py2status'] = 'released'
+    for pkg in pkg_by_version[2]:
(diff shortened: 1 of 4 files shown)


data/statuses.yaml
```diff
@@ -37,6 +37,26 @@
...
+- ident: py2done
+  name: Py2Done
+  abbrev: ''
+  color: ''
+  term: ''
+  weight: 7
+  description: Python 2 subpackage is correctly named
+  rank: 7
+  instructions: |
+    The Python 2 subpackage follows the naming policy.
```
encukou: Could this be done with the usual `mispackaged` and `released` statuses? It seems they wouldn't clash, since these are in a different field. Or is there a special reason to add new ones?

irushchyshyn: It is possible.
Their description though is very Py3 oriented, but I could probably just edit the descriptions.

encukou:
Looks good, thanks!

PR #277, opened 2017-04-11
Track misnamed packages (Step 2 for #268)

irushchyshyn:
Step 2 for #268.

- [x] Mark as `misnamed` packages like sympy (where there is `foo` and `python3-foo`);
- [x] Rename tracking attribute to `is_misnamed` and make it boolean;
- [x] Store name status as an attribute of RPM rather then the whole package, and display them with a warning badge;
- [x] Add UI view for naming policy tracking.

Index view:
photo_2017-04-11_11-10-08
Going to `Misnamed` link:
photo_2017-04-11_11-10-04
Package view:
screenshot from 2017-04-11 11-23-32
screenshot from 2017-04-11 11-24-39

Any feedback will be appreciated.


hroncok:
I like the way it is displayed. Will have a look at the code soon (today).

dnf-plugins/py3query.py
```diff
@@ -174,16 +174,31 @@ def set_status(result, pkgs, python_versions):
...
+
+        # Missing python2- prefix (e.g. foo and python3-foo).
+        missing_prefix = (
+            'python' not in pkg.name and
+            # There is a python3-foo package.
+            ('python3-{}'.format(pkg.name) in name_by_version[3] or
+             '{}-python3'.format(pkg.name) in name_by_version[3]) and
+            # There is no python2-foo package.
+            ('python2-{}'.format(pkg.name) not in name_by_version[2] and
+             '{}-python2'.format(pkg.name) not in name_by_version[2]))
```
hroncok: Can the above code has a function? if you need a comment to explain something, maybe replace it with more easy to grasp function? Also DRY.

(untested)

```python
def has_pythonX_package(pkg, name_by_version, version):
   """Checks whether pythonX-foo or foo-pythonX exists"""
    return 'python{}-{}'.format(version, pkg.name) in name_by_version[version] or
￼           '{}-python{}'.format(pkg.name, version) in name_by_version[version]

...

missing_prefix = (
    'python' not in pkg.name and
    has_pythonX_package(pkg, name_by_version, 3) and
    not has_pythonX_package(pkg, name_by_version, 2)
)

```

portingdb/htmlreport.py
```diff
@@ -729,6 +747,49 @@ def mispackaged():
...
+
+    # Statuses with number of packages
+    statuses = OrderedDict(
+        db.query(tables.Status, func.count(tables.Package.name))
+        .outerjoin(tables.Status.packages)
+        .join(tables.CollectionPackage)
+        .filter(tables.CollectionPackage.collection_ident == 'fedora')
+        .filter(tables.CollectionPackage.is_misnamed.is_(True))
+        .group_by(tables.Status.ident)
+        .order_by(tables.Status.order))
```
hroncok: JFY it hard to grasp this query for me as I'm not so familiar with the data structure, so I assume it works. Ask @encukou for the review here if you feel like it needs one.

portingdb/tables.py
```diff
@@ -227,16 +219,16 @@ class CollectionPackage(TableBase):
     nonblocking = Column(
         Boolean(), default=False,
         doc=u"If true, does not block dependent packages (even if it's marked as unported)")
+    is_misnamed = Column(
+        Boolean(), doc=u"True if the package does not follow the naming policy")
```
hroncok: I tough this is Python 3 only, but I see that other docs also have `u""`, so keep it to be consistent.

dnf-plugins/py3query.py
irushchyshyn: Yes, this should work and will look better. Thanks for a suggestion, will change it.

portingdb/tables.py
irushchyshyn: Right.

portingdb/htmlreport.py
irushchyshyn: I requested a review from @encukou  too.

It would be easier to query this if Package would also have `is_misnamed` attribute, but I do not want to store it in three different tables.

encukou: In the beginning of this project, I made a normalized relational database. But as the project grew, I realized it's only edited by dropping everything and re-loading from the data in Git, so normalization doesn't really make sense.

Please add a column to the database if it makes your job easier. The DB is effectively a cache.
That said no reason to add columns until needed. If you already wrote code without the column, keep it.

As for the review, could you ping me once @hroncok's comments are addressed?

irushchyshyn: Thanks for explanations. I already made it work without the column. This query looks complicated with all the joins, but it does the job, so I'll keep it. 
The rest of the comments are addressed.

hroncok: @encukou ping

portingdb/htmlreport.py
```diff
@@ -158,6 +168,15 @@ def get_groups(db, query):
     return groups
 
 
+def get_naming_progress_info(correct_count, misnamed_count):
+    NameStatus = namedtuple('NameStatus', 'name color term url description')
```
encukou: Please define classes outside of functions, unless there's a reason not to do it.

portingdb/templates/namingpolicy.html
```diff
@@ -0,0 +1,55 @@
...
+    <div class="col-md-12">
+        {{ progress_summary(progress, total_packages) }}
+    </div>
+    <div class="col-md-12">
+        <p>
+            This is a dashboard to track packages which do not comply with 
+            <a
+               href="https://fedoraproject.org/wiki/Packaging:Naming?rd=Packaging:NamingGuidelines#Python_source_package_naming" 
+               target="_blank">Python package naming guidelines</a>.
+            These packages are to follow the new guidelines before 2020 / Fedora 29,
```
encukou: Fedora releases twice a year, so in 2020 we should have Fedora 32 (or 33).

portingdb/htmlreport.py
irushchyshyn: Fixed in 72ed03b

portingdb/templates/namingpolicy.html
irushchyshyn: As we do not know the release number yet, I just removed it and only left the year (in 72ed03b)

encukou:
Looks good!
Please add a regular "update Fedora data" commit to this PR, so the JSON is updated with the code.

irushchyshyn:
Thanks a lot for the review!

Updated Fedora data:

**mispackaged -> released (1)**

- python-ngram
besser82, Björn Esser, [commit](https://src.fedoraproject.org/cgit/rpms/python-ngram.git/commit/python-ngram.spec?id=a363dd3e588a49dec440cf20ccd55f3d6205f285)

PR #279, opened 2017-04-20
Collect data on unversioned requires

irushchyshyn:
Add `misnamed_deps` to be collected for each package in `fedora.json` if the package uses unversioned names in requirements (e.g. `python-foo`).  This will help to keep track of packages which will potentially have broken dependencies when the switch for `python3` to be `python` is done.

According to the initial results, there are 1279 packages, which use wrong requires, 1043 of which are blocked by misnamed packages (those, not providing python2-foo).

### DIFF
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -282,6 +282,19 @@ class Py3QueryCommand(dnf.cli.Command):
                 if req in python_versions.keys():
                     deps_of_pkg[req].add(pkg)
 
+        wrong_requirers = self.pkg_query.filter(
+            requires__glob=['python-*', '[!/]*-python-*', '[!/]*-python'])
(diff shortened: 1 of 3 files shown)


encukou:
I think this should be done for *all* packages in Fedora. It's not very useful to have it only for the packages in portingdb.

irushchyshyn:
@encukou , what do you mean by all packages? I thought that if the package has at least one Python package in requirements, then we track it in portingdb.

encukou:
We only track packages that depend *directly* on Python (or GTK bindings, but that's a special case).

irushchyshyn:
In 949a906 I have made it to collect unversioned requirers within all packages in Fedora and store as `unversioned_requirers` for each package in fedora.json.

According to this data, there are 1412 packages with incorrect requires, and 1343 of them are tracked in portingdb.

encukou:
Looks good, thanks!

PR #286, opened 2017-05-09
Display unversioned requirers on Naming Policy page

irushchyshyn:
This PR contains UI for  #279 and #283.

Index view:
image

Naming policy view:
image

### DIFF
--- a/data/naming.yaml
+++ b/data/naming.yaml
@@ -1,28 +1,45 @@
 - ident: name-correct
   name: Correct
-  color: 'd9d9d9'
+  color: '669980'
   term: '✔'
-  description: Correctly named packages
-  misnamed: false
+  short_description: All good
+  description: All good
+  violation: All good
 - ident: name-misnamed
   name: Misnamed
   color: 'F0AD4E'
   term: '×'
-  misnamed: true
-  url: namingpolicy
-  description: Packages that do not follow the naming policy
-- ident: require-correct
-  name: Correct
-  color: 'd9d9d9'
-  term: '✔'
-  description: All requirements are versioned
-- ident: require-misnamed-blocked
-  name: Misnamed
-  color: 'F0AD4E'
+  violation: |
+    using the unversioned naming scheme in binary RPMs, which is against the
+    new <a href="https://fedoraproject.org/wiki/Packaging:Naming?rd=
+    Packaging:NamingGuidelines#Python_source_package_naming"
+    target="_blank">Python package naming guidelines</a>
+  short_description: Naming scheme of (at least) one of the binary RPMs is wrong
+  description: |
+    The naming scheme of (at least) one of the binary RPMs
+    does not comply with the naming policy. Please refer to the
+    <a href="https://python-rpm-porting.readthedocs.io/en/latest/naming-scheme.html">
+    Python RPM Porting Guide</a> for more information and instructions on the changes
+    needed.
+- ident: require-unversioned
+  name: Unversioned Requires
+  color: '8c6699'
   term: '×'
-  description: Some requirements are not versioned because no versioned name is provided
-- ident: require-misnamed-not-blocked
+  violation: |
+    using the unversioned naming scheme in run-time or build-time requirements,
+    although a versioned name is available
+  short_description: The package uses wrong naming scheme in requires
+  description: |
+    Some build-time or run-time requirements are not versioned,
+    although a versioned name is available (can be fixed now).
+- ident: require-misnamed
   name: Blocked
-  color: 'F0AD4E'
+  color: 'a65959'
   term: '×'
-  description: Some requirements are not versioned although a versioned name is provided
+  violation: |
+    using the unversioned naming scheme in run-time or build-time requirements,
+    because the requirement is misnamed and the versioned name is not provided
+  short_description: The package depends on a misnamed package
+  description: |
+    Some build-time or run-time requirements are not versioned, because no
+    versioned name is available (waiting for the misnamed package to be renamed).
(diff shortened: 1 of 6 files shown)


hroncok:
I like how it looks. However I have a bad feeling about "unversioned requires", in my mind, that means the package requires `python2-requests` instead of `python2-requests >= 2.2`; not `python-requests` vs. `python2-requests`. Can we call it "Misnamed requires" instead?

irushchyshyn:
Good point @hroncok , thanks! I will change it.

data/naming.yaml
```diff
@@ -21,25 +21,30 @@
...
   color: '8c6699'
   term: '×'
   violation: |
-    using the unversioned naming scheme in run-time or build-time requirements,
-    although a versioned name is available
-  short_description: The package uses wrong naming scheme in requires
+    some build-time or run-time requirements are of type <i>python-foo</i>,
+    although the names with a versioned Python prefix
+    (e.g. <i>python2-foo</i>) are available
+  short_description: The package uses wrong naming scheme in Requires
```
hroncok: (Build)Requires maybe?

irushchyshyn: Amended to 14f4a06

encukou:
Nice job! I have a few nitpicks regarding the presentation:

* [ ] The Non-Python packages link to the spec file. That's a useful link, but it's not the only useful one; it would be nice to link to the Fedora Package Database as well (e.g. [kajongg](https://src.fedoraproject.org/cgit/rpms/kajongg.git/tree/kajongg.spec)).
Maybe you can use a table, like the one for Mispackaged packages on the main page, and add the two links in columns?

* [ ] The small "progress bar" on the main page (below Naming Policy in the sidebar) is hard to read: the naming policy colors are too dark for contrast with black text. 
It should be OK to remove the text from all of these progress bars; it's not really useful in the other ones.

* [ ] In the same sidebar, the fields with explanations need the little color boxes to visually link them to the progressbar.

* [ ] I'm not sold on using the Naming policy colors for the vertical bars of color before the package names. Could you try to use the regular status colors, even here? That color tends to follow the package across portingdb.
  * [ ] If you decide to keep the naming policy colors: the vertical bar should always be on the same line as the package name. (For regular status colors the macro should ensure this.)

One thing I wanted to ask: what happens if a package has both a **Misnamed Subpackage** and **Misnamed Requires**?

irushchyshyn:
Thanks a lot for the review @encukou !

>     The Non-Python packages link to the spec file. That's a useful link, but it's not the only useful one; it would be nice to link to the Fedora Package Database as well (e.g. kajongg).
>     Maybe you can use a table, like the one for Mispackaged packages on the main page, and add the two links in columns?

That would be useful, I agree. I will try adding the table and will see how it looks.

>     The small "progress bar" on the main page (below Naming Policy in the sidebar) is hard to read: the naming policy colors are too dark for contrast with black text.
>     It should be OK to remove the text from all of these progress bars; it's not really useful in the other ones.

Agreed, will remove the text.

>     In the same sidebar, the fields with explanations need the little color boxes to visually link them to the progressbar.

I will fix it.

>     I'm not sold on using the Naming policy colors for the vertical bars of color before the package names. Could you try to use the regular status colors, even here? That color tends to follow the package across portingdb.

I thought it would be a bit confusing, but I will try to keep the original colors there and we will see.

> One thing I wanted to ask: what happens if a package has both a Misnamed Subpackage and Misnamed Requires?

Good question, I thought of it for a while too. In this case, the package will go into `Misnamed Subpackage` category, as it seems more important to unblock the packages which depend on it. Only after the name is fixed, it will got into `Misnamed Requires` category, if it still has the issue. Do you have another idea about it?

encukou:
> > One thing I wanted to ask: what happens if a package has both a Misnamed Subpackage and Misnamed Requires?
> 
> Good question, I thought of it for a while too. In this case, the package will go into `Misnamed Subpackage` category, as it seems more important to unblock the packages which depend on it. Only after the name is fixed, it will got into `Misnamed Requires` category, if it still has the issue. Do you have another idea about it?

That implies a workflow I'd rather avoid: us opening a bug for *Misnamed Subpackage*, waiting for the maintainer to fix, and then opening a bug for the *Misnamed Requires* – even though they could both be fixed in one update.

Could you just note this “overriding” on the Naming Policy Violations page? The raw data is there, we need to keep this in mind for a mass bug filing script, but I don't think you need to spend time on visualising it perfectly.

irushchyshyn:
Of course, I will add a note.

irushchyshyn:
Review updates are available in 4a80cc1:

- [x] Removed text from the small progress bars;
- [x] Added little color boxes for Naming policy sidebar legends to link them to the progress bar:
image
- [x] Used regular status colors for vertical bars before the package names. Shown in the screen below, does it look better?
- [x] Added a note on "overriding" the Misnamed Requires status for packages  with misnamed subpackages:
image
- [x] For Non-Python Packages, also added a link to Fedora Package DB, and displayed as a table:
image

hroncok:
> Used regular status colors for vertical bars before the package names. Shown in the screen below, does it look better?

Not for my OCD :D

irushchyshyn:
I tend to agree. So I decided to keep the naming policy colors, but fixed the vertical bar position. Amended to  b0c1d75

>  If you decide to keep the naming policy colors: the vertical bar should always be on the same line as the package name. (For regular status colors the macro should ensure this.)

encukou:
I've added a fix for one more nitpick (directly to the PR, to save a review round-trip).
If you agree, let's merge this.

irushchyshyn:
Thanks @encukou , it looks good.

PR #289, opened 2017-05-15
wxPython is Python 3 upstream available

ari3s:


encukou:
Thanks! This'll appear in the next update.

PR #298, opened 2017-06-01
DNF plugin: Mark packages that just have py2 binaries as nonblocking

encukou:
Packages such as python-pbr have Python 2 binaries, but are otherwise
ported. Portingdb currently shows them as mispackaged, so anything
that depends on them shows up as Blocked, even though they can be
ported.

Mark these packages as nonblocking.


irushchyshyn:
👍 
Should we merge it and then update the data?

encukou:
Yes, or if you can just pull this before the next update, it would work as well.

PR #300, opened 2017-06-06
Packages Python 3 upstream available

ari3s:
- [x] limnoria [[https://pypi.python.org/pypi/limnoria](https://pypi.python.org/pypi/limnoria)]
- [x] python-sphinxcontrib-httpdomain [[https://pypi.python.org/pypi/sphinxcontrib-httpdomain](https://pypi.python.org/pypi/sphinxcontrib-httpdomain)]

### DIFF
--- a/data/groups.yaml
+++ b/data/groups.yaml
@@ -353,7 +353,6 @@ django:
     - python-django-countries
     - python-django-database-url
     - python-django-debug-toolbar
-    - python-django-devserver
     - python-django-discover-runner
     - python-django-dynamite
     - python-django-extensions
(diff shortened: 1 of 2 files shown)


irushchyshyn:
Opened [Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1459474) for python-sphinxcontrib-httpdomain.

Regarding limnoria, the spec file says that [not all plugins are ported](https://src.fedoraproject.org/cgit/rpms/limnoria.git/tree/limnoria.spec#n2), and that is the reason it's Python 2 so far. So not opening a bug for now.

PR #309, opened 2017-06-27
Weak dependencies for naming policy

### LINKED ISSUE #307: Naming policy does not seem to consider Recommends
http://fedora.portingdb.xyz/namingpolicy/ does not seem to consider `Recommends`. I haven't checked other soft dependency types.

For example, the git-bz package has `Recommends: python-crypto` at the moment, meaning the Python 2 version. python2-crypto is available, but git-bz is not listed in the naming policy page as a violator.

irushchyshyn:
- Check weak dependencies for naming policy
- Ignore Python 3 only packages when checking dependencies
- Ignore unqualified dependency on documentation packages (python-foo-doc)

Fixes #307

### DIFF
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -254,13 +254,11 @@ class Py3QueryCommand(dnf.cli.Command):
                 if req in python_versions.keys():
                     deps_of_pkg[req].add(pkg)
 
-        unversioned_wildcards = ['python-*', '[!/]*-python-*', '[!/]*-python']
-        runtime = self.pkg_query.filter(requires__glob=unversioned_wildcards)
-        buildtime = self.src_query.filter(requires__glob=unversioned_wildcards)
         # unversioned_requirers: {srpm_name: set of srpm_names}
         unversioned_requirers = collections.defaultdict(set)
-        for pkg in progressbar(runtime + list(buildtime), 'Getting unversioned requirers'):
-            for require in pkg.requires + pkg.requires_pre:
+        for pkg in progressbar(self.all_unqualified_requires(), 'Getting unversioned requirers'):
+            for require in (pkg.requires + pkg.requires_pre + pkg.recommends +
+                            pkg.suggests + pkg.supplements + pkg.enhances):
                 require = str(require).split()[0]
                 requirement = all_provides.get(require)
                 if is_unversioned(require) and requirement:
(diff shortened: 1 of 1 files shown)


dnf-plugins/py3query.py
```diff
@@ -366,3 +364,24 @@ def whatrequires(self, dep):
         query = self.pkg_query
         query = query.filter(requires=dep)
         return set(query)
+
+    def all_unqualified_requires(self):
+        """Return a set of packages, which have misnamed
+        runtime, buildtime or weak dependencies.
+        """
+        unqualified_wildcards = ['python-*', '[!/]*-python-*', '[!/]*-python']
+        requireres = set()
```
encukou: Typo: `requireres`

dnf-plugins/py3query.py
```diff
@@ -254,13 +254,11 @@ def run(self, args):
...
-        unversioned_wildcards = ['python-*', '[!/]*-python-*', '[!/]*-python']
-        runtime = self.pkg_query.filter(requires__glob=unversioned_wildcards)
-        buildtime = self.src_query.filter(requires__glob=unversioned_wildcards)
         # unversioned_requirers: {srpm_name: set of srpm_names}
         unversioned_requirers = collections.defaultdict(set)
-        for pkg in progressbar(runtime + list(buildtime), 'Getting unversioned requirers'):
-            for require in pkg.requires + pkg.requires_pre:
+        for pkg in progressbar(self.all_unqualified_requires(), 'Getting unversioned requirers'):
+            for require in (pkg.requires + pkg.requires_pre + pkg.recommends +
+                            pkg.suggests + pkg.supplements + pkg.enhances):
```
encukou: This all takes quite a lot of time. Could you print a message about that, so the plugin doesn't look stuck?

dnf-plugins/py3query.py
irushchyshyn: Fixed

dnf-plugins/py3query.py
irushchyshyn: `self.all_unqualified_requires()` takes a lot of time. I have added info messages for querying each type of dependencies, and the log now looks as follows:
```
...
[====================] 3/3 Querying packages with unversioned dependencies: Done!     
[====================] 3906/3906 Processing packages with unversioned dependencies: Done!
...
```

encukou:
Looks good, thanks!

PR #315, opened 2017-07-21
py3query.py update

irushchyshyn:
After update to F26 the `py3query.py` plugin started to fail with:
`dnf py3query: error: unrecognized arguments: -o data/fedora.json`

- [x] Update the plugin structure according to the latest [dnf version recommendations](http://dnf.readthedocs.io/en/latest/use_cases.html#plugins-cli-api)
- [x] Remove `if self.opts.help_cmd` conditionals, which I believe were fixed upstream in [commit](https://github.com/rpm-software-management/dnf/commit/24345fe8dd8223fb83cc65d4607c5222d4d55ab2)
- [x] Update docs to not be bound to python3.5

### DIFF
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -10,7 +10,6 @@ This will give you a file "fedora.json" with information for portingdb.
 
 from __future__ import print_function
 
-import operator
 import sys
 import json
 import collections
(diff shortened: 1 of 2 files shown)


encukou:
Unfortunately, this doesn't work with f25's dnf any more.
I'm fine with requiring the latest version, but could you add a note that f26+ is needed?

encukou:
Alright, I'll trust you that it works :)

I've made a `f25` branch hers to mark the version that still works on the old system.

irushchyshyn:
Perfect, thanks!

PR #317, opened 2017-07-25
check-specs script: Check if packages are built for epel

irushchyshyn:
A follow up for #312
```
Packages that use %{?rhel} in spec files: 571 of 2338 (24.42%)
  name-misnamed: 240
  require-misnamed: 165
  require-blocked: 166
From those 571, have el6 or epel7 branch: 465
```

### DIFF
--- a/data/rhel-specs.json
+++ b/data/rhel-specs.json
@@ -1,0 +1,1838 @@
+{
+  "name-misnamed": {
+    "NLopt": [
+      "el6",
+      "epel7"
+    ],
+    "OpenImageIO": [
+      "el6",
+      "epel7"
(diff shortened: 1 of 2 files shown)


encukou:
Could you make it output a list of the mackages in JSON for further analysis?
Also, since it takes a while to run this, could you put that result in `data/`? Include a timestamp in the file if it's not regenerated on every update.

encukou:
And, could you also include the EPEL info in that file? Plus possibly note in the docs how to run the tool to generate the file.

irushchyshyn:
Sure. Amended the changes to e71a246

PR #318, opened 2017-07-25
Mark packages that depend on unversioned /usr/bin/python

encukou:
Add a marker like this:

page-shot-2017-7-25 beakerlib python 3 porting database

I don't think a full tracking page is necessary – a simple repoquery is enough to get a list of these packages.

### DIFF
--- a/portingdb/templates/package.html
+++ b/portingdb/templates/package.html
@@ -76,6 +76,12 @@
                                                     <span class="fa fa-li"> </span>
                                                 {% endif %}
                                                 {{ pydep.name }}
+                                                {% if pydep.name == '/usr/bin/python' %}
+                                                    <span class="badge"
+                                                        style="background-color: #F0AD4E !important;"
+                                                        title="Scripts should explicitly use /usr/bin/python3 or /usr/bin/python2, especially in shebangs"
+                                                    >⅔</span>
+                                                {% endif %}
                                             </li>
                                         {% endfor %}
                                     </ul>


irushchyshyn:
Looks good!
What about `usr/bin/env python`?

And do you think it would make sense to add similar handling for `python-*` dependencies on the package page?

hroncok:
> What about `usr/bin/env python`?

Since we are looking at requires and not the actual shebangs, we cannot get those. The require would be just `/usr/bin/env`. We can display that one as well with either ⅔ or a question mark.

encukou:
Before starting to track `/usr/bin/env` in portingdb, I'd like to gather some more info. Specifically, are all/most uses of `/usr/bin/env python*` in Fedora wrong (because they should always use the Python provided by the system? Or do we only need to get rid of the unversioned `/usr/bin/env python`?

Anyway, the solution to that will most likely be very different, so I'd rather not combine it with this PR.

hroncok:
> Specifically, are all/most uses of `/usr/bin/env python*` in Fedora wrong (because they should always use the Python provided by the system?

Yes. And I was even sure there is guideline about that, but apparently, it was never made an actuall guideline, although, there is a rpmlint check.

https://fedoraproject.org/wiki/Script_Interpreters_(draft)
https://fedoraproject.org/wiki/Features/SystemPythonExecutablesUseSystemPython
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/WYBBLP2PVCVVJVKBDVXDQMJRQ4AT4P35/

PR #319, opened 2017-07-25
py3query: Gather build dependencies

encukou:
For wrangling misnamed packages/requires, it would be useful to also track build dependencies in portingdb. Here's a patch that adds the info to `fedora.json` (not to the DB or UI yet).
I wrote this on the f25 branch and then cherry-picked; hopefully it will also work on master / Fedora 26.

### DIFF
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -215,7 +215,7 @@ class Py3QueryCommand(dnf.cli.Command):
 
             for dep in progressbar(provides, 'Getting py{} requires'.format(n)):
                 dep_versions[str(dep)] = n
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
well. build dependencies are complicated. SRPM is arch-dependent while Fedora publishes only one, random one... =(

encukou:
I think it's OK if portingdb doesn't capture all the complexity. The
existing data also leaves out a lot :)

I imagine this to be used mostly to answer questions like "how many
packages depend on this one - 5, or 500?", or to build a graph to plan a
mass bug filing strategy. For specific, individual queries, things like dnf
repoquery will always be better.

PR #322, opened 2017-07-26
py3query.py: Track /usr/bin/env dependencies

encukou:
This is a bit of a hack -- the field is named "py_deps" but is now used to
track all specific dependencies (as opposed to SRPMs) that are useful to us.

Again, I tested the DNF plugin on f25, and cherry-picked. Please test on f26 too, before merging.

### DIFF
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -268,7 +268,7 @@ class Py3QueryCommand(dnf.cli.Command):
         # get extra dependencies
         extra_deps = ['/usr/bin/env']
         for dep in progressbar(extra_deps, 'Getting extra dependencies'):
-            for pkg in self.whatrequires(dep):
+            for pkg in self.whatrequires(dep, self.pkg_query):
                 if pkg in python_versions.keys():
                     rpm_pydeps[pkg].add('/usr/bin/env')
 
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -205,7 +205,7 @@ class Py3QueryCommand(dnf.cli.Command):
         # rpm_pydeps: {package: set of dep names}
         rpm_pydeps = collections.defaultdict(set)
         # dep_versions: {dep name: Python version}
-        dep_versions = collections.defaultdict(set)
+        dep_versions = collections.defaultdict(int)
         for n, seeds in SEED_PACKAGES.items():
             provides = sorted(self.all_provides(reponame, seeds), key=str)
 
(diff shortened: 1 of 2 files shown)


dnf-plugins/py3query.py
```diff
@@ -265,6 +265,13 @@ def run(self):
                         pkg.sourcerpm).name if pkg.sourcerpm else pkg.name
                     unversioned_requirers[requirement_srpm_name].add(requirer_srpm_name)
 
+        # get extra dependencies
+        extra_deps = ['/usr/bin/env']
+        for dep in progressbar(extra_deps, 'Getting extra dependencies'):
+            for pkg in self.whatrequires(dep):
```
irushchyshyn: Please provide `self.pkg_query` as a second argument to ` self.whatrequires`.

irushchyshyn:
The message would be misleading if the package has  `/usr/bin/env <not python>` shebangs.

For example, 389-ds-base
screenshot from 2017-07-26 16-36-12
which has only `usr/bin/env perl` shebangs:
```
ishcherb@ishcherbina:~/Downloads/test$ grep -r  /usr/bin/env .
./usr/lib64/dirsrv/perl/DSCreate.pm:    my $perlexec = "/usr/bin/perl" || "/usr/bin/env perl";
./usr/bin/logconv.pl:#!/usr/bin/env perl
./usr/bin/cl-dump.pl:#!/usr/bin/env perl
```

We should either make the message more general, or find a way to figure out which `/usr/bin/env` it is. I do not know an easy way to do it, except greping the files in rpm.

encukou:
I plan to reuse code from taskotron-python-versions for this, after https://github.com/fedora-python/taskotron-python-versions/issues/28 is implemented there

encukou:
With shebang mangling, this is no longer relevant. I ran the code and found out only about 70 packages we track still rely on `/usr/bin/env`:

```
blender
blktrace
blueman
ceph-base
ceph-common
ceph-mon
dmlite-dpm-tester
dmlite-shell
docker-unit-test
icaro
dpm-python
libstoragemgmt-arcconf-plugin
libstoragemgmt-hpsa-plugin
libstoragemgmt-local-plugin
libstoragemgmt-megaraid-plugin
libstoragemgmt-netapp-plugin
libstoragemgmt-nfs-plugin
libstoragemgmt-nstor-plugin
libstoragemgmt-smis-plugin
libstoragemgmt-targetd-plugin
python3-libstoragemgmt
lirc-config
lirc-core
matrix-synapse
mediawiki
nordugrid-arc-arex
ntop
nut-client
olpc-switch-desktop
pcs
python2-django-flash
python2-pivy
python2-twisted
python2-pytz
qpid-qmf-devel
qpid-tools
sagemath-notebook
sagemath-sagetex
varnish-devel
veusz
wesnoth-data
xen-runtime
```

I no longer think it make sense to track this in portingdb.

PR #325, opened 2017-08-01
Add the `py3-only` status

encukou:
This adds an indication of which packages shouldn't be affected when Python 2 is dropped.


encukou:
I think it'd be better in the next update – to check if it works on another machine :)

PR #337, opened 2017-09-14
Proof of Concept: Freeze with Elsa

hroncok:
How to freeze:

   * create venv
   * do what README says (except don't `pip isntall --user`, but in venv)
   * `pip install elsa`
   * `python elsasite.py freeze --serve`
   * go get a coffee

Note that ideally, freezing would be utilized with `python -m portingdb freeze`, but that is blocked by https://github.com/pyvec/elsa/pull/39 and that one is stalled for now.

If we go this ways, we might also want to get rid of dogpile.

Don't forget to: document this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 __pycache__/
 /*.egg-info/
+/portingdb/_build
 /portingdb.sqlite
 /portingdb.sqlite-shm
 /portingdb.sqlite-wal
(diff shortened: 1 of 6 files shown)


hroncok:
This freezes!

hroncok:
Testing on Travis CI.

hroncok:
The job exceeded the maximum time limit for jobs, and has been terminated.

So, not on Travis. But local freeze works :)

irushchyshyn:
We can deploy it ourselves, right?

hroncok:
Yes.

PR #342, opened 2017-09-30
Add upstream updates for a few packages.

troycurtisjr:
Add upstream info for mintlocale and openbox.  Update upstream info for alacarte.

### DIFF
--- a/data/upstream.yaml
+++ b/data/upstream.yaml
@@ -298,6 +298,15 @@ m2crypto:
 meld:
   links:
     homepage: http://meldmerge.org/
+mintlocale:
+  status: in-progress
+  links:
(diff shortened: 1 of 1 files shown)


encukou:
Thank you!
I've pushed this as 8ab48aabd15d31d4e73df594d6a88e29919e6d03 with minor modifications:
- Not describing current status, as that tends to get out of sync. Instead, describe what has heen done.
- Not using the `in-progress` state, which also tends to get out of sync

The current notes don't follow this – best practices here are kind of evolving :)

encukou:
This will be live after the next update (which might be delayed somewhat, we switched hosting recently).

troycurtisjr:
Looks good, thanks.  I'll keep that in mind for any future PRs I have. 👍

PR #348, opened 2017-10-31
Update Fedora data

irushchyshyn:
**idle -> py3-only (1)**
- cinnamon-desktop
leigh123linux, Leigh Scott, badges 1->2 (wiki updated)

**missing -> idle (2)**
- macromilter, blocked by python-pymilter
- speedometer, does not work with Py3 [currently](https://src.fedoraproject.org/rpms/speedometer/blob/master/f/speedometer.spec#_8)

**missing -> py3-only (1)**
- nitroshare

**missing -> released (6)**
- petsc4py
- python-backports-csv
- python-dominate
- python-feedgen
- python-jnius
- python-oletools

**_Naming status changes_**

**missing -> requires blocked (1)**
- speedometer

**missing -> requires misnamed (2)**
- python-backports-csv
- i-nex

**requires misnamed -> requires ok (2)**
- fedpkg
- rpkg

**requires ok -> requires blocked (1)**
- python-ldap3


encukou:
This is neat!

Do you use a script for Naming status changes? If so, could you put it in `scripts/` or update the one that's already there?

irushchyshyn:
@encukou I did update the `scripts/jsondiff.py`. But it is a little messy now and I would like to polish it a bit before pushing it, so I will create a PR for it soon.

encukou:
+1

PR #355, opened 2017-11-30
Use Ambiguous Requires instead of Misnamed Requires on UI 

irushchyshyn:
Also add a cli command to easily get a list of packages.

image


portingdb/cli.py
```diff
@@ -437,3 +438,23 @@ def closed_mispackaged(ctx):
 
     if results:
         exit(1)
+
+
+@cli.command()
+@click.option(
+    '-l', '--list-packages',
+    type=click.Choice(['misnamed-subpackage', 'ambiguous-requires', 'blocked']))
```
encukou: Options are optional. Use [`click.argument`](http://click.pocoo.org/5/arguments/) instead.

encukou:
Thanks! This does sound more friendly and to the point. I merged the first commit.

portingdb/cli.py
irushchyshyn: Right, thanks. Fixed.

PR #367, opened 2018-01-10
ROS group

ari3s:


data/groups.yaml
```diff
@@ -516,3 +516,22 @@ trac:
     - trac-watchlist-plugin
     - trac-workflowadmin-plugin
     - trac-xmlrpc-plugin
+ros:
+    name: ROS
```
encukou: I think it should be named e.g. `Packages from the Robot Operating System`, so people who don't know what [ROS](https://en.wikipedia.org/wiki/Reactive_oxygen_species) is can find out.

PR #375, opened 2018-02-01
Show year numbers in progress graphs

encukou:
Graphs with month names are getting quite crowded now that portingdb is past its second year.

Use year numbers for progress graphs that show 2+ years; year numbers + short month names otherwise.


irushchyshyn:
Looks good! Thanks!
But we might bump into this by June:
image
 Is there any way to maybe show the year a bit lower, when the months are also shown?

encukou:
Probably, but I've pushed month names to the right instead. It should be readable until it switches to years only.

irushchyshyn:
Looks better, thanks :) 👍 

image

PR #381, opened 2018-02-28
Remove the `in-progress` status and the "Where porting happens" sidebar

encukou:
These seemed like good ideas at the time, but weren't.
The in-progress status stays in history; it was used until late 2016.

### DIFF
--- a/portingdb/htmlreport.py
+++ b/portingdb/htmlreport.py
@@ -53,12 +53,6 @@ def hello():
     query = queries.packages(db)
     total_pkg_count = query.count()
 
-    active = query.filter(tables.Package.status == 'in-progress')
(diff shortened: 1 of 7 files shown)


portingdb/htmlreport.py
```diff
@@ -138,13 +131,11 @@ def hello():
         breadcrumbs=(
             (url_for('hello'), 'Python 3 Porting Database'),
         ),
-        collections=collections,
         coll_info=coll_info,
```
irushchyshyn: `coll_info` is also not used.
And would be nice to remove `collections` and `coll_info` variables from the function as well.

encukou:
Unfortunately I won't have much time to devote to portingdb now, so please do it in a future update.

Issue #382, opened 2018-03-01
Track Build Requires


irushchyshyn:
Track also packages that require Python or any Python package to build.

Create a new status (e.g "legacy leaf") that would display Released packages, whose python2 subpackages are not needed by anything.

 ~~Those are SRPMs which depend on one of [SEED_PACKAGES](https://github.com/fedora-python/portingdb/blob/master/dnf-plugins/py3query.py#L52). Add those packages to the database, and display SRPM python requirements on the package page above RPMs.~~

This is an initial overview of the requirement. @encukou let me know if you have other ideas regarding how they should be viewed.

encukou:
We should rather track all packages that BuildRequire *any* python package (or at least, any py2 package). It could be done the same way as [non-Python pkgs for misnamed stuff](http://fedora.portingdb.xyz/namingpolicy/#require-misnamed-non-python), i.e. just use simple names rather than a full-fledged entry.

The final outcome in portingdb should be to show "legacy leaf" packages, i.e. `released` ones whose py2 subpackages aren't needed for anything else in Fedora. (And that should be tracked as a new status.)

irushchyshyn:
I see now, thanks for clarifying it for me. I will start with data collection for that.

encukou:
I believe BuildRequires are tracked well now. @irushchyshyn, if you had any more improvements in mind, could you re-open this?

irushchyshyn:
Sure. The only thing we could add is displaying non-Python `Dependent packages` on the package page, but this is not of high priority. They are taken into account when identifying Python 2 leaves, but just are not displayed.

PR #383, opened 2018-03-02
Track legacy leaf Python 2 packages among released

irushchyshyn:
Actually we have started to collect BuildRequires of Python packages in #319.
What was left is tracking any non-Python packages that require portingdb packages.

Scope:
- [x] For each RPM in a package, collect non-Python packages which depend on it (both build time and runtime). This adds a new field to `fedora.json` in the `rpms` section:

```
"non_python_requirers": {
          "build_time": [],
          "run_time": []
        },
```

- [x] Add a new status `legacy-leaf`, which means that all Python 2 RPMs in a package are not required by anything in Fedora (that RPMs will have a `legacy_leaf: True` field). This status will be applied only to `Released` packages (those that package both Python versions of the package).
- [x] Add a new status to the index page on UI (between Released and Py3 only)
Index page top:
image
Legacy-leaf section:
image

- [x] Display which RPM is a `legacy_leaf`
image

Further optional:
(Not to make this PR even bigger, these are to be done in separate PRs)
- [ ] Display BuildRequires on package page
- [ ] Display non-Python requirers on package page

---
<details>
  <summary>Fedora data update</summary>

**released -> legacy-leaf (710)**
- APLpy
- ATpy
- COPASI
- GeographicLib
- NLopt
- adapt
- ansible-inventory-grapher
- ansible-review
- argparse-manpage
- assimp
- atomic-reactor
- aubio
- botan
- bpython
- brltty
- cairo-dock-plug-ins
- cantor
- capstone
- caribou
- catfish
- clufter
- clustershell
- collectd-ceilometer-plugin
- compose-utils
- conda
- conu
- cryptsetup
- csdiff
- dlib
- dnf-plugins-extras
- dnfdaemon
- etckeeper
- fence-agents
- firewalld
- flann
- fontdump
- gcc-python-plugin
- gdcm
- gerrymander
- ginga
- git
- glances
- gpaw
- gprof2dot
- gpsd
- gtts
- hivex
- i2c-tools
- ibus
- imgbased
- iscsi-initiator-utils
- isomd5sum
- jabberpy
- keycloak-httpd-client-install
- lcm
- ldns
- libcap-ng
- libforensic1394
- libfreenect
- libftdi
- libkml
- liblinear
- liblouis
- libnl3
- libnuml
- libproxy
- libsbml
- libsedml
- libsolv
- libssh2-python
- libstoragemgmt
- libyui-bindings
- link-grammar
- mathgl
- module-build-service-copr
- moose
- mraa
- nbdkit
[...]

PR #386, opened 2018-03-06
Load build dependencies into the database

irushchyshyn:
* On the package page, categorize dependencies into build time and run time
image

* Add relationships for build and run time dependencies separately:
`package.run_time_requirements`, and `package.build_time_requirements`

So far build dependencies are only being shown on the package page left panel in `Dependent packages` and `Dependency tree` to showcase. They can also be displayed in the `Dependencies` sections, on Groups page etc. Also only run time dependencies are so far favored when deciding which packages are Blocked or when building a dependency graph.

Scope of #382

### DIFF
--- a/portingdb/cli.py
+++ b/portingdb/cli.py
@@ -170,12 +170,12 @@ def report(ctx):
     query = db.query(tables.Package)
     query = query.order_by(func.lower(tables.Package.name))
     query = query.options(eagerload(tables.Package.by_collection))
-    query = query.options(subqueryload(tables.Package.requirements))
+    query = query.options(subqueryload(tables.Package.run_time_requirements))
     for package in query:
         print_collection_info(package, collections)
         print(' ' + package.name, end=' ')
         reqs = []
-        for req in package.requirements:
+        for req in package.run_time_requirements:
             for cp in req.by_collection.values():
                 if cp.status not in ('released', 'legacy-leaf', 'py3-only'):
                     reqs.append(req.name)
@@ -262,7 +262,7 @@ def deps(ctx, package, exclude, trim, skip, graph):
             e = '✔'
         else:
             seen.add(pkg)
-            reqs = [p for p in pkg.requirements if p is not pkg]
+            reqs = [p for p in pkg.run_time_requirements if p is not pkg]
             if skip:
                 reqs = [r for r in reqs
                         if not (can_ignore(r) or r.name in exclude)]
--- a/portingdb/htmlreport.py
+++ b/portingdb/htmlreport.py
@@ -210,6 +210,15 @@ def package(pkg):
 
     dependents = list(queries.dependents(db, package))
 
+    query = queries.build_dependencies(db, package)
+    query = query.options(eagerload('status_obj'))
+    query = query.options(subqueryload('collection_packages'))
+    query = query.options(subqueryload('collection_packages.links'))
+    query = query.options(eagerload('collection_packages.status_obj'))
+    build_dependencies = list(query)
+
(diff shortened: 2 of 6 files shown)


irushchyshyn:
Note:
The numbers on the top progress bar have slightly changed in comparison to current state of portingdb:
image

The reason for that, is after we introduced collecting ambiguous requirements, we were adding some build only ambiguous requirements to the package dependencies (which is a bug). Therefore e.g. [koji](http://fedora.portingdb.xyz/pkg/koji/) in protingdb depends on python-spinx, however it is a build time dependency only. I have fixed this in this PR, thus the inconsistencies.

encukou:
This conflicts with #383 so I'll hold off the review, but I can comment on your comments:

> So far build dependencies are only being shown on the package page left panel in Dependent packages and Dependency tree to showcase. They can also be displayed in the Dependencies sections, on Groups page etc.

It would be great to consider them in *determining* Group membership. And when that's done, show them in the Group view.
But that can be a separate commit.

The package page left panel is not immediately necessary, but you can add it for completeness' sake.

> Also only run time dependencies are so far favored when deciding which packages are Blocked or when building a dependency graph.

I think that's fine.

encukou:
Looks good at first glance. I probably won't have time to review in the next few days, feel free to double-check and merge.
Or if you feel something needs extra scrutiny, please point it out.

irushchyshyn:
I am not sure about a couple of things:
1) On the Group page we show all the packages needed for the group (no separation between build time and runtime deps). I did it because the dependency lists get too long and not clear, if I do build time + run time categories as on the Package page. Is it ok like this, or do we need to add categories for some special use cases?

2) Do we need to track packages that build require Python, but do not build any RPMs that require Python? We currently do not. We only track build requirements of packages, which already existed in portingdb.

3) More general question and probably not much related to this PR.
Python 2 leaf packages are currently determined while running the dnf Py3Query (not the way we determine e.g `blocked` packages). This is mainly due to 2 reasons:
* we save `SRPM to SRPM` deps into database (not `RPM to RPM`, or `SRPM to RPM`)
* we do not save non-Python (non in portingdb) dependencies at all (except for a couple of ambiguous ones).
So, I am not sure if it is fine like that, or do I have to think about storing dependencies differently?

encukou:
> On the Group page we show all the packages needed for the group (no separation between build time and runtime deps). I did it because the dependency lists get too long and not clear, if I do build time + run time categories as on the Package page. Is it ok like this, or do we need to add categories for some special use cases?

That should be fine. Or we can tweak the display in the future.

> Do we need to track packages that build require Python, but do not build any RPMs that require Python? We currently do not. We only track build requirements of packages, which already existed in portingdb.

If we start tracking these, I guess they'd just have an empty list of RPMS.
But I don't think it's necessary *right* now.

> More general question and probably not much related to this PR.
> Python 2 leaf packages are currently determined while running the dnf Py3Query (not the way we determine e.g blocked packages). This is mainly due to 2 reasons:
>
> 1. we save SRPM to SRPM deps into database (not RPM to RPM, or SRPM to RPM);
> 2.  we do not save non-Python (non in portingdb) dependencies at all (except for a couple of ambiguous ones).
>
> So, I am not sure if it is fine like that, or do I have to think about storing dependencies differently?

The main thing is that determining leaf packages takes the built requirements into account – they're leaves within Fedora, not within portingdb's set of tracked packages.
I wouldn't bother changing the mechanism until we have a plan of what to do with the new data.

irushchyshyn:
@encukou thanks for clarifying everything.
I do not have anything else here, so I am merging this PR. If anything would seem wrong, feel free to file a bug.

PR #387, opened 2018-03-08
Follow up corrections for issue #382

irushchyshyn:
- [x] On index page subqueryload only run time requirements (as it was before);
- [x] Do a bit of refactoring for Py3Query:
Skip the step of querying ambiguous requires, as after we started collecting all build and run time requirers of python packages, we already have the data needed.
- [x] Add `legacy-leaf` status for history data collection;
- [x] Do Fedora update (although there were no composes) to add `legacy-leaf` status to `history.csv` and make sure the query refactoring did not break anything.

#382


encukou:
Thanks!

PR #395, opened 2018-03-21
 Set correct ContentType for CSV files 

encukou:
This is a useful part of #337. Creating a PR to see if Travis is happy.


encukou:
Oops, no Travis

PR #398, opened 2018-03-23
Small docs fixes and single script for update

irushchyshyn:
I have updated the doc for portingdb slightly, and added a simple script that does all the steps for the update (it is a bit ugly, if you use a better one, feel free to replace it).

Also removed `check-specs.py`, which has served it's purpose and I do not think it will ever be used again. Plus, it uses pkgdb, which has been decommissioned.


encukou:
It looks good! Let me iterate on it when I find possible improvements :)

PR #401, opened 2018-03-26
3 changes

encukou:
Three unrelated changes:

* Allow a `groups-update` file
* Remove a mention of Fedora in the main template
* Don't celebrate dual py2/py3 support

Sending as one PR to cut down on bureaucracy, but I can split them if one is controversial.

### DIFF
--- a/portingdb/templates/index.html
+++ b/portingdb/templates/index.html
@@ -157,7 +157,7 @@
             <h2 id="legacy-leaf">Packages with leaf Python 2 subpackages</h2>
             <div>
                 The Python 2 versions of {{ len(legacy_leaf_packages) }} packages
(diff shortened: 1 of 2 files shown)


portingdb/templates/index.html
```diff
@@ -147,7 +147,8 @@ <h2 id="blocked">Blocked packages</h2>
             </ul>
             <h2 id="released">Packages with dual support</h2>
             <div>
-                {{ len(released_packages) }} packages support both Python 2 and 3. Yipee!
+                {{ len(released_packages) }} packages support both Python 2 and 3.
+                Half-way there!
```
hroncok: like_a_sir

PR #403, opened 2018-03-29
Option to load both run and build time deps on the graph

### LINKED ISSUE #402: Build time dependency graph
We seem to track build time dependencies now. See how [asciidoc](http://fedora.portingdb.xyz/pkg/asciidoc/) has plenty build time dependent packages. Yet in our [dependency graph](http://fedora.portingdb.xyz/graph/) there is no big cluster around it.

Could we please have a second, build-time dependency (possibly combined with runtime) graph?

irushchyshyn:
Fixes #402 

Default behavior when opening graph page:
image

Clicking on "load build time deps as well" (see asciidoc cluster)
screenshot from 2018-03-29 16-50-18

### DIFF
--- a/portingdb/htmlreport.py
+++ b/portingdb/htmlreport.py
@@ -375,16 +375,18 @@ def graph_json(grp=None, pkg=None):
             package = todo.pop()
             if package not in requirements:
                 requirements.add(package)
-                todo.update(p for p in package.run_time_requirements
-                            if p.status not in DONE_STATUSES)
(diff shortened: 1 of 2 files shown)


encukou:
Nice!
The graph is so small that I think it would be fine to always show all deps. But an option is great too!

irushchyshyn:
> The graph is so small that I think it would be fine to always show all deps.

That would make it all more simple :) We can make it a default if there are no use cases for run time deps only :)

PR #419, opened 2018-05-04
Make the history graph nicer

encukou:
Yesterday at a Python meetup there was a talk about Plotly, a library for interactive graphs.
I had been ignoring it so far because the company behind it prefers that you use a web API, but I was surprised to learn you can use it for "offline" graphs as well... and they're quite committed to open source!

I'm preparing a talk about the switch to py3, where I'll use the graph, so I decided to modernize it a bit. 
What do you think?

newplot

This PR also changes the status colors a bit.


portingdb/history_graph.py
```diff
@@ -0,0 +1,95 @@
...
+    'mispackaged',
+    'idle',
+    'blocked',
+)
+
+def order_key(status_ident):
+    try:
+        return STATUS_ORDER.index(status_ident)
+    except ValueError:
+        return 4
```
irushchyshyn: What means 4?:)

encukou: It puts any unknown status somewhere between "ported" and "unported". I left that grossly underdocumented, sorry! (but the actual value not that important)

encukou:
It's not intentional – I just put in the colors and left the styling to the experts :)
It should be possible to style this with CSS, but if it's just a nitpick I'll not look into it further now.

PR #427, opened 2018-06-11
holland now supports Python 3

carlwgeorge:


encukou:
Thanks for the note! It should go live in the next update.

PR #436, opened 2018-08-16
python-caja is done

carlwgeorge:


data/badges.txt
```diff
@@ -697,6 +697,7 @@ tomspur Thomas Spura
 carlwgeorge Carl George
     https://src.fedoraproject.org/rpms/keychecker/c/7d193856505bdf6364acce92873332c47645bab9?branch=master
     https://src.fedoraproject.org/rpms/evemu/c/5e7d5d8ed77030dbba64127638f2f9caa95890fd
+    https://bugzilla.redhat.com/show_bug.cgi?id=1560724
```
encukou: The entry doesn't include the component name, so we can't use it as it is. No need to fix it on your side, though – the entry should be recorded in the next update.
I'll keep the PR open until that's done.

carlwgeorge: Ah yeah I see your point, sorry about that.  I don't mind fixing it.

encukou:
Done as part of a normal update.

PR #440, opened 2018-09-07
Check which packages whould be dropped

encukou:
The check-drops subcommand decides what packages to drop for the  [mass py2 package removal change](https://fedoraproject.org/wiki/Changes/Mass_Python_2_Package_Removal).
Now that the change is accepted, it makes sense to have it in portingdb proper.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ __pycache__/
 /portingdb.sqlite-shm
 /portingdb.sqlite-wal
 *.pyc
+/_check_drops/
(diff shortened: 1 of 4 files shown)


encukou:
A merge would preserve history without additional work. But if rebasing works better for you, go for it – I don't mind at all!

encukou:
ACK for the README, but you might find that file needs a larger update.
Just mentioning the script and pointing to its docstring is probably enough.

PR #450, opened 2018-10-02
Bypass the SQL database and freeze with Elsa

encukou:
Putting the relational DB in portingdb was a mistake in the beginning:
- We don't update the data once loaded, so we can build "indexes"
- Converting DB rows ↔ Python objects is so slow we needed a cache anyway
- Python is a much nicer syntax for data manipulation than SQL
- All the data easily fits in memory, and always will

... and, well, I now know a lot more about building sites that are only updated with pull requests.

This adds a way to load all the needed data to in-memory dicts, and switches the bulk of the site to use that instead of SQL.
Some things (CLI, Naming status, graphs) still use the database.
Some things I've just removed.

The result is fast enough to freeze in a reasonable time, so we can host this on GH Pages.


hroncok:
How urgent is this? Can it wait for Thursday?

encukou:
It can definitely wait.
In the mean time, I'm building better reports for py2 dropping on top of this. I can rebase those when it's merged.

hroncok:
The "Dependent packages" list in package detail is not sorted by status.

hroncok:
"Package Groups" and "Other Groups" used to be sorted lexicographically.

encukou:
Here I'm unlinking the "build time dependencies" graphs and "mispackaged" with patch requested. Those aren't very useful nowadays.
The percentages graph is moved to its own URL.

"Package Groups" and "Other Groups" are sorted.

For sorting "Dependent packages" and "Dependency Tree", I'd like to do a few more changes first.

encukou:
Try this: info on dependent packages in the same way as the dependencies.
Sorry for putting a new feature in this PR, but it's easier than sorting the "Dependent packages"/"Dependency Tree" lists only to revert that later.

hroncok:
The Ui now feels a bit weird, but I guess it's just that I'm not used to it. Would collapsible sections make it better?

I.e. after the overview section:

Run time 
16 packages:
 11 Dual support;
 5 Dropped
XXX see the list XXX

and the (sometimes long) list would roll down...


Also, I've noticed this tooltip:

encukou:
> The Ui now feels a bit weird, but I guess it's just that I'm not used to it. Would collapsible sections make it better?
> 
> I.e. [...] the (sometimes long) list would roll down...

Myself, I'd rather scroll.
How bad is this for you? Can we merge, and perhaps discuss UX in person later?

> Also, I've noticed this tooltip:

Fixed, thanks.

encukou:
Tomorrow, if time permits, I'll assume it's bikeshedding, merge and send a PR that builds on top of this.
Let me know if you want more time.

PR #452, opened 2018-10-07
Update Fedora data

encukou:


hroncok:
**python2-prompt-toolkit** should be marked as dropped (by **python-prompt-toolkit**, that produces **python3-prompt-toolkit**).

Issue #459, opened 2018-10-11
python2-networkx is not tracked


hroncok:
Can we somehow add python2-networkx package manually?

See https://bugzilla.redhat.com/show_bug.cgi?id=1634570#c3

encukou:
Hmm. It's not a Python package:
```
dnf repoquery --repo=rawhide -l python2-networkx
/usr/share/doc/python2-networkx
/usr/share/doc/python2-networkx/README.rst
/usr/share/licenses/python2-networkx
/usr/share/licenses/python2-networkx/LICENSE.txt
```
Obviously the "almost-leaf" thing is not perfect. I'm not sure how to best fix this, however; I'll need to think/experiment a bit.

For a workaround, do the same thing as for python2-matplotlib – keep the subpackages manually. That's better than adding python2-networkx, IMO.

hroncok:
It's a python2 metapackage. It exists just to bring in python2-networkx-xxx.

encukou:
Also, python2-networkx was removed by the maintainer...

PR #460, opened 2018-10-12
Track subpackages with "python" in the name

encukou:
#  Track all subpackages with a "python" component in the name

Track all subpackages with `python2`, `python3` or `python` as a
dash-separated component of the name.
For subpackages that don't have a Python dependency (as found by
the existing mechanism), mark them `python2` or `python3`
according to the names. (If it's just `python`, don't mark it
either way.)

This will find a lot of new "misnamed requires".

<s>This should cover `python2-networkx` and similar metapackages, so it
fixes: https://github.com/fedora-python/portingdb/issues/459
</s>

 ## Update Fedora data to pick up plugin changes

**missing -> idle (3)**
- check-mk
- python-mtTkinter
- python2-docs

**missing -> released (8)**
- libdnet
- openmpi
- python-basemap-data
- python-repoze-what
- python-rpm-macros
- python-typeshed
- python3-docs
- texlive

**released -> idle (3)**
- python-docker
- python-oslo-config
- python-pyface

**released -> legacy-leaf (1)**
- nautilus-python

**_Naming status changes_**
**missing -> requires misnamed (12)**
- gedit-plugins
- jython
- pypy
- pypy3
- python34
- python35
- python36
- fawkes
- backintime
- python-rpm-macros
- redhat-rpm-config
- texlive

**requires ok -> requires misnamed (51)**
- ibus
- prepaid-manager-applet
- gcc-python-plugin
- gstreamer-rtsp
- libldb
- livecd-tools
- nbdkit
- python-afl
- python-basemap
- python-cartopy
- python-collectd_systemd
- python-formencode
- python-gencpp
- python-genlisp
- python-genpy
- python-ipykernel
- python-metakernel
- root
- python-multilib
- python-music21
- python-oslo-cache
- python-oslo-concurrency
- python-oslo-db
- python-oslo-log
- python-oslo-middleware
- python-oslo-policy
- python-oslo-privsep
- python-oslo-utils
- python-oslo-versionedobjects
- python-oslo-vmware
- python-virtualenv
- python2
- python3
- pyside-tools
- python-qt5
- swatchbooker
- copr-rpmbuild
- ecryptfs-utils
- jpype
- pulp-rpm
- python-aiosmtpd
- python-atpublic
- python-cov-core
- python-flufl-bounce
- python-flufl-i18n
- python-flufl-lock
- python-flufl-testing
- python-iso8601
- python-nose2
- python-ryu
- python-sphinx

### DIFF
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -244,6 +244,21 @@ class Py3QueryCommand(dnf.cli.Command):
                     python_versions[pkg].add(n)
                     rpm_pydeps[pkg].add(str(dep))
 
+        # Add packages with 'python?' as a component of their name, if they
+        # haven't been added as dependencies
+        for name, version in {
+            'python': 0,
+            'python2': 2,
+            'python3': 3,
+        }.items():
+            for pattern in '{}-*', '*-{}', '*-{}-*':
+                name_glob = pattern.format(name)
+                query = self.pkg_query.filter(name__glob=name_glob)
+                message = 'Getting {} packages'.format(name_glob)
+                for pkg in progressbar(query, message):
+                    if pkg not in python_versions:
+                        python_versions[pkg].add(version)
+
         # srpm_names: {package: srpm name}
         # by_srpm_name: {srpm name: set of packages}
         srpm_names = {}
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -256,6 +256,9 @@ class Py3QueryCommand(dnf.cli.Command):
                 query = self.pkg_query.filter(name__glob=name_glob)
                 message = 'Getting {} packages'.format(name_glob)
                 for pkg in progressbar(query, message):
+                    if pkg.sourcerpm.startswith('mingw-'):
+                        # Ignore mingw packages
+                        continue
                     if pkg not in python_versions:
                         python_versions[pkg].add(version)
 
--- a/data/fedora.json
+++ b/data/fedora.json
@@ -492,7 +492,9 @@
       "libevent",
       "python3"
     ],
-    "deps": [],
+    "deps": [
+      "python3"
+    ],
     "links": {
       "bug": [
         "https://bugzilla.redhat.com/show_bug.cgi?id=1323243",
@@ -526,6 +528,7 @@
     "deps": [
       "python-AppTools",
       "python-envisage",
+      "python-pyface",
       "python-traitsui",
       "python2",
       "vtk"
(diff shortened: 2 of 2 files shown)


encukou:
@hroncok, what do you think about this approach?

hroncok:
Looking at this.

hroncok:
Should we be worried about the mingw stack?

hroncok:
> This will find a lot of new "misnamed requires".

Taskotron check uses a whitelist:

https://github.com/fedora-python/taskotron-python-versions/blob/16ca63862428ec1012e8ec3f79db977291bd14a1/taskotron_python_versions/two_three.py#L35

Doesn't portingdb use that code? Or is it just that latest release is soo old?

dnf-plugins/py3query.py
```diff
@@ -244,6 +244,21 @@ def run(self):
...
+            'python': 0,
+            'python2': 2,
+            'python3': 3,
+        }.items():
+            for pattern in '{}-*', '*-{}', '*-{}-*':
+                name_glob = pattern.format(name)
+                query = self.pkg_query.filter(name__glob=name_glob)
+                message = 'Getting {} packages'.format(name_glob)
+                for pkg in progressbar(query, message):
+                    if pkg not in python_versions:
```
hroncok: What about `and not pkg.stratswith('mingw')`?

encukou:
> Taskotron check uses a whitelist. Doesn't portingdb use that code?

Not for this part. Taskotron is used to check py2-only binaries and naming policy.

Taskotron doesn't really apply to this PR. This is about the subset of Fedora's packages that portingdb tracks; Taskotron runs on everything.

hroncok:
I meant for the missnamed requires. Some of them are already tracked in taskotron as "not missnamed". Reusing that would give us less false missnamed requires hits.

encukou:
Rebased onto #461, the latest data update.

Meanwhile, the networkx maintainer dropped all python2 subpackages anyway :/

encukou:
> I meant for the missnamed requires. Some of them are already tracked in taskotron as "not missnamed". Reusing that would give us less false missnamed requires hits.

That's for another issue. I'm not worrying about misnamed packages right now; there's a lot of other stuff going on.

dnf-plugins/py3query.py
hroncok: AFAIK the binary packages are called mingw32- or mingw64-

hroncok: oh, it's sourcerpm.startswith, nice!

PR #464, opened 2018-10-18
Check Drops improvements

hroncok:


portingdb/check_drops.py
```diff
@@ -228,6 +229,7 @@ def handle_entrypoints(result, config):
                          'tw2.widgets', # plugins for a framework, not app
                          'gearbox.project_commands',
                          'gearbox.commands',
+                         'openstack.data_processing.v1',
```
hroncok: From python2-saharaclient.

Issue #414, opened 2018-04-19
RFE: Summary by packager


hroncok:
I've juts found out that one of my packages (openscad) requires python2 by accident.

I would like to check all of my packages and having a page at portingdb with my packages would be very helpful. See https://pagure.io/fedora-misc-package-utilities/blob/master/f/find-package-maintainers for how to get the info.

frenzymadness:
I don't know portingdb internals but IMHO this todo list might describe this RFE implementation:
- [ ] Add a new table to handle packagers' logins with foreign key to packages
- [ ] Add a way to download/update JSON file from pagure.io
- [ ] Add a new section to `load_from_directories()` which loads data from JSON file to database
- [ ] Add a new CLI command to check "my packages"
- [ ] Add a new web page with info about all packages and their maintainers with some JS search/filter

What do you think? If this is correct, I can try to implement it.

encukou:
You can skip the database, just load a dict from JSON.
I'm trying to (slowly) get rid of the DB.
See the load_data module.

hroncok:
https://src.fedoraproject.org/extras/pagure_owner_alias.json

hroncok:
I'd say:

 1. Download that json when updating data, store it in the data folder
 2. Crete a page that shows a detail for one maintainer (similarly how groups do it)
 3. Create a page with a sorted list of maintainers with links to above

encukou:
Instead of your 3., I went with:

3. Link to maintainer pages from package pages

frenzymadness:
Thank you! Now I can drop all python2- subpackages from packages of mine.

hroncok:
Only the ones that nothing depends on, please.

PR #468, opened 2018-10-23
Add maintainer pages

encukou:
The maintainer page is only linked (and thus frozen) for maintainers of 2+ packages.

Fixes: https://github.com/fedora-python/portingdb/issues/414


portingdb/templates/_base.html
```diff
@@ -195,6 +195,20 @@
     {% endif %}
 {%- endmacro %}
 
+{%- macro maintainer_link(maintainer) -%}
+    {% if maintainer.packages | length > 1 %}
+        <a href="{{ url_for('maintainer', name=maintainer.name) }}">
+            {{ maintainer.name }}
+        </a>
```
hroncok: this makes the space part of the link, maybe we can make this one line instead:

     <a href="{{ url_for('maintainer', name=maintainer.name) }}">{{ maintainer.name }}</a>

encukou: Jinja has dashes for that :)

PR #480, opened 2018-11-23
Stop tracking py3 deps; rename to "Python 2 Dropping Database"

encukou:


portingdb/templates/index.html
```diff
@@ -4,7 +4,9 @@
 {% endblock breadcrumbs %}
 
 {% block titlecontent -%}
-    {{ (the_score * 100) | round(1) }}% – {{ super() }}
+    {{ (the_score * 100) | round(1) }}% →
+    {{ (py2_score * 100) | round(1) }}% –
+    {{ super() }}
```
hroncok: The new title is a bit comprehensive to grasp.

data/statuses.yaml
```diff
@@ -26,7 +26,7 @@
 - ident: released
   name: Dual support
   abbrev: "\u2714"
-  color: '7EBA54'
+  color: '5EBA84'
```
hroncok: This one feels particularly ugly to my eyes. But I understand that this is not objective feedback.

encukou:
> Why is "update data" the first commit? Does it reflect the new thing?

GitHub orders the commits based on date. That's not the order I intended – I did some rebasing :(
Could you check that locally?

hroncok:
Oh, GitHub... :D

portingdb/templates/index.html
encukou: So, I tried explaining it on the main page.

data/statuses.yaml
encukou: Design changes do that. If it still looks jarring in a week, let's change it?

portingdb/templates/index.html
```diff
@@ -24,42 +64,68 @@ <h1>
         <div class="col-md-12">
             {{ progress_summary(status_summary, total_pkg_count) }}
         </div>
+        <div class="col-md-12">
+            <div class="main-summary">
+            <div class="summary-sentence">
+                <span class="big-percentage"
+                    style="color: {{ statuses['py3-only'].color }};"
+                >
+                    {{ (the_score * 100) | int }}%
```
hroncok: round? at least the title now says 44.9 %, but the page says 44 %.

portingdb/htmlreport.py
```diff
@@ -56,19 +74,17 @@ def hello():
 
     by_status = group_by_status(packages.values())
 
-    the_score = sum(len(by_status[s]) for s in DONE_STATUSES) / len(packages)
+    the_score = len(by_status['py3-only']) / len(packages)
+    py2_score = sum(len(by_status[s]) for s in DONE_STATUSES) / len(packages)
```
hroncok: dropped is in DONE_STATUSES...

portingdb/templates/index.html
```diff
@@ -24,42 +64,68 @@ <h1>
...
+                </span>
+                <span class="stuffing">
+                    Python packages in {{ config['name'] }} support
+                </span>
+                <span class="callout">Python 3 only.</span>
+            </div>
+            <div class="summary-sentence">
+                <span class="big-percentage"
+                    style="color: {{ statuses['released'].color }};"
+                >{{ (py2_score * 100) | int }}%</span>
```
hroncok: ...this includes dropped packages and hence is not true as in "X% Python packages in Fedora support Python 3."

portingdb/templates/index.html
encukou: Alright, let me fit that in

portingdb/templates/index.html
encukou: Good point. The goal changed now; "dropped" is not really "done". I'm keeping `DONE_STATUSES` as they were and adding a new set.

encukou:
Ooh, nice! I just noticed GitHub links to a diff after I force-push.

encukou:
Thanks!

PR #486, opened 2018-12-03
Avoid DB in the naming policy & history

encukou:
Switch Naming policy to use data in dicts, rather than going through the DB.

There are some changes to how naming policy is calculated: mainly, Misnamed packages no longer block themselves. Also, instead of a single status, a package can both be Misnamed and have Ambiguous Requires, and "Blocked" is a subset of Ambiguous Requires.

There are also changes to how naming policy is displayed: mainly, the status shows up as a colorful box on nearly any mention of a package in the HTML report.


hroncok:
This is huge! Individual commits make sense, let me check the result.

encukou:
Thanks for the review!

PR #487, opened 2018-12-03
Add FTBFS warnings throughout the HTML report

encukou:
This is by no means rigorous, but still a nice indication.

### DIFF
--- a/portingdb/load_data.py
+++ b/portingdb/load_data.py
@@ -15,6 +15,10 @@ from .load import _merge_updates
 PY2_STATUSES = {'released', 'legacy-leaf', 'py3-only'}
 DONE_STATUSES = PY2_STATUSES | {'dropped'}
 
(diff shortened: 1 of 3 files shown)


portingdb/load_data.py
```diff
@@ -15,6 +15,10 @@
 PY2_STATUSES = {'released', 'legacy-leaf', 'py3-only'}
 DONE_STATUSES = PY2_STATUSES | {'dropped'}
 
+# Current Fedora version, used to get info about long-term FTBFSs
+# Update to the Rawhide version after a mass rebuild.
+CURRENT_FEDORA = 29
```
hroncok: We might advice to do it a month or two after a mass rebuild not to have the portingdb overwhelmed by this.

encukou: Just use common sense when updating it :)

encukou:
I don't intend this to be anything more than a quick heads-up that the package has problems.

PR #488, opened 2018-12-05
Statscache is decommissioned

frenzymadness:


hroncok:
@cverna Since the repo is archived and I cannot comment there: Any plans to retire the package(s) from Fedora?

cverna:
> @cverna Since the repo is archived and I cannot comment there: Any plans to retire the package(s) from Fedora?

Sure, I have opened https://pagure.io/fedora-infrastructure/issue/7421 so someone with the proper permissions can do it.

encukou:
Is there a python2-less alternative to this software? If there's none, it shouldn't be marked `dropped`. File a bugzilla to retire it in Fedora.

---
More info on where that came from:

Where possible, portingdb aims to be a *dashboard* rather than a *data source*.
The `dropped` status is not "the official directory of old cruft *someone* wants to be gone" – that would be unmaintainable. If the maintainer wants it to be gone, they should orphan/retire it; anyone else (including Python-SIG/python-maint) needs to work with the maintainer.
Instead, `dropped` marks things with a clear py2-less alternative, like python2 (use python3!), yum (use dnf!) or `python-enum` (use enum in py3 stdlib!). This info is there primarily for maintainers of the *dependent* packages. That data is not anywhere else, and we need it (without it, everything would be `blocked` by python2). We're a data source in this particular case, but have a clear scope to avoid devolving into "the directory of old cruft".

There are currently 3 exceptions. I'm not particularly happy about them, but can manage – handling them better is one of my TODO items, whose priority will increase if things start spiraling out of control. (Actually, it shouldn't be on my personal TODO list. I filed https://github.com/fedora-python/portingdb/issues/490.)
If you come up with a maintainable process to manage "things someone wants gone", let's start tracking them (preferably Fedora-wide). But `dropped` is not it.

PR #491, opened 2018-12-06
Make the JS graph show Python 2 dropping status

encukou:
Also:
- Remove Group graphs entirely
- Avoid the DB in the web app
- Require networkx

And, this makes the graph CPU-hungry again.

### DIFF
--- a/portingdb/templates/graph.html
+++ b/portingdb/templates/graph.html
@@ -22,13 +22,13 @@
     }
 
     .link {
-    stroke: #bbb;
-    stroke-opacity: 1;
+    stroke: #000;
+    stroke-opacity: 0.1;
     }
 
     .marker {
-    fill: #bbb;
-    fill-opacity: 1;
+    fill: #000;
+    fill-opacity: 0.1;
     }
 
     .d3-tip {
@@ -90,7 +90,6 @@ var width = 1200,
     height = 900;
 
 var force = d3.layout.force()
-    .linkDistance(15)
     .size([width, height]);
 
 var svg = d3.select("#graph-goes-here").append("svg")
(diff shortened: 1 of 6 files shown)


hroncok:
Not sure how do I read that graph yet, but it is indeed CPU hungry :D

hroncok:
Shouldn't the text above the graph be updated?

encukou:
Ah, good catch! I knew I forgot something.

portingdb/templates/graph.html
```diff
@@ -47,31 +47,13 @@
...
-                <a href='{{ url_for("package", pkg=pkg) }}'>{{ pkg }}</a>
-                package, as
-            {% endif %}
-            tracked by <a href='{{ url_for("hello") }}'>portingdb</a>.
-            Packages with no relationship to other unported packages are not
-            shown.
-            Darker colors mean more dependencies.
+            Here is a graph showing dependency relationships for
+            packages in Fedora Rawhide that need Python 2 in some way.
+            Brighter colors roughly mean the package can be removed sooner.
```
hroncok: OK, I guess I don't get that part. I see the same colors as in portingdb. How is their brightness showing when they can be removed?

encukou: The ones that can't be removed easily are dark (and they're in the center).
Granted, it's not very obvious – and perhaps not even helpful. Should this just use the same colors as everywhere else?
Here's the graph with this shading turned off:

btightgraph

hroncok: I only see it on the green ones. And only now when you showed me. Better ask another person as well.

encukou: Well, the "hard to remove" ones *are* mostly green (and gray).

encukou: Are we bikeshedding at this point?

hroncok: Yes. Keep the colors.

PR #504, opened 2019-01-09
Update Fedora data

encukou:
**idle** → **missing** (3)
- gst-inspector
- kcbench-data
- zbar

**legacy-leaf** → **released** (2)
- python-streamlink
- python-stuf

**mispackaged** → **idle** (1)
- python-muranoclient

**missing** → **idle** (1)
- sugar-story (FTBFS fix, python2 dep is not actually new)

**missing** → **released** (2)
- python-bsddb3
- python-tvb-gdist

**released** → **legacy-leaf** (3)
- python-iso-639
- python-iso3166
- python-parse

**released** → **missing** (2)
- python3-bsddb3
- singularity

## Naming status changes

**named correctly** → **missing** (5)
- gst-inspector
- kcbench-data
- python3-bsddb3
- singularity
- zbar

**requires misnamed** → **missing** (1)
- heimdal

---

No history update – the script picks up the first commit each day, which had no change.

### DIFF
--- a/data/fedora.json
+++ b/data/fedora.json
@@ -188,7 +188,7 @@
     ],
     "deps": [],
     "rpms": {
-      "Carla-2.0.0-0.7.20181225git2f3a442.fc30": {
+      "Carla-2.0.0-0.8.20181225git2f3a442.fc30": {
         "almost_leaf": false,
         "legacy_leaf": false,
         "non_python_requirers": {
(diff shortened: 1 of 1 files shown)


hroncok:
Ha. python-muranoclient went from mispackaged to idle, because the bug is closed; but it is closed because I've retired python-muranoclient.

I guess the data is not consistent at this point. Do we care? It will be solved with the next update after a new rawhide compose.

PR #505, opened 2019-01-10
Refresh groups a bit, mostly Xfce

hroncok:
I was looking at compos, but it got tedious, not sure if this can be somehow automated instead?

### DIFF
--- a/data/groups.yaml
+++ b/data/groups.yaml
@@ -2,11 +2,13 @@ identity:
   name: Identity Management
   hidden: false
   packages:
+    - 389-ds-base
     - freeipa
     - samba
     - ipsilon
+    - sssd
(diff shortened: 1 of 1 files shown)


hroncok:
Compos could be the data source to use but there are several challenges:

 * the groups would need to be resolved recursive as they can contain other groups
 * the groups would need to be hand picked, as there is too many
 * the groups contain binary packages, not source packages
 * some packages in the group are not tracked in portingdb but can require stuff that is tracked in portingdb (I consider this the most challenging bit.)

To summarize: a nice introductory challenge for a newcomer, but nether you or me would invest their time into this.

encukou:
OK, probably not worth automating.
Thanks for the update, though!

PR #506, opened 2019-01-10
Remove some out-of-date notes

encukou:
portingdb started out as a collection of notes, but now that it's mostly aggregating data tracked elsewhere, the notes are unnecessary and usually obsolete.
I removed some old data during an offline train ride.

This PR is not complete.


encukou:
I don't really want to drop it all at once without looking at it. But looking at it isn't a big priority.

PR #518, opened 2019-02-18
Mark Fedora 30 as current to display Mass Rebuild FTBFS

encukou:
There's a lot of FTBFS markers … except in the py3-only packages. I think it is useful to have this.
@hroncok, please take a look (after your vacation) & let me know if it looks useful.


hroncok:
Let's give it 4 more weeks.

hroncok:
Give or take 1 day, that's now.

PR #519, opened 2019-02-22
Include python2-requests-cache for MythTV

hobbes1069:
As requested in:

https://bugzilla.redhat.com/show_bug.cgi?id=1629795#c10


hroncok:
Thank You, Richard!

PR #525, opened 2019-03-26
Remove the SQL database code

encukou:
The data is small enough to be kept in dicts.


hroncok:
load is no more. it needs to be removed form Travis and docs.

encukou:
Thanks!

PR #536, opened 2019-05-21
Allow to specify repo for check drops

hroncok:
I have no better idea how to change click default / required of one option based on another option's provided value.


encukou:
Make the default `None`, and update it if you get None. Would that work?

hroncok:
Will try.

hroncok:
Amended.

encukou:
Looks good! I didn't try it yet though.

hroncok:
It works OK for me so far.

PR #537, opened 2019-05-21
Mark python2-more-itertools dropped

hroncok:
See https://bugzilla.redhat.com/show_bug.cgi?id=1582171
and https://bugzilla.redhat.com/show_bug.cgi?id=1712501


encukou:
Thank you!

PR #546, opened 2019-07-08
 Don't track dependencies on version-agnostic wheels

encukou:
Explicitly don't track dependencies on `python-pip-wheel`, `python-setuptools-wheel`, `python-wheel-wheel` – portingdb looked at these names and thought they might be old-style Python2 packages.
(Note that we only track python2 dependencies.)

This untangles some loops in the build dependency graph.

---
\+ 1 small unrelated fix

### DIFF
--- a/data/fedora.json
+++ b/data/fedora.json
@@ -6903,9 +6903,7 @@
     "status": "py3-only"
   },
   "copr-rpmbuild": {
-    "build_deps": [
-      "python-rpm-macros"
-    ],
+    "build_deps": [],
     "deps": [],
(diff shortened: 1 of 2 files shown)


hroncok:
I thought we track those here: https://github.com/fedora-python/taskotron-python-versions/blob/49a81c51ee9c583fa3e0c9e2501c77451bac64f4/taskotron_python_versions/two_three.py#L35 - do you want a new release?

encukou:
Ah, right. I forgot we track those externally. Sorry for the noise!

Issue #549, opened 2019-07-19
FTBFS packages


frenzymadness:
Packages qmtest and pacrunner seem to fail to build from source in Fedora for some time but they are not marker properly in portingdb.

Bugs lists:
* https://bugzilla.redhat.com/buglist.cgi?quicksearch=qmtest&list_id=10347096
* https://bugzilla.redhat.com/buglist.cgi?quicksearch=pacrunner&list_id=10347095

encukou:
portingdb only has a FTBFS *warning*, a heuristic based on the built packages' names – it flags anything ending with `.fcX` with X < 30. It's not meant to be comprehensive.

`pacrunner` built for f30 successfully.
`qmtest` has [minorbump](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_you_need_to_change_an_old_branch_without_rebuilding_the_others), which this doesn't account for, but it's probably not worth the trouble to fix it in portingdb. (But feel free to send a PR if you disagree.)

frenzymadness:
That's all I need to know, thank you!

PR #550, opened 2019-07-19
Consider minorbump when evaluating dist tag

hroncok:
Packages with release like 2.fc29.1 will now be marked as FTBFS.

See https://github.com/fedora-python/portingdb/issues/549

I actually look at packages with multiple explanation marks and try to "nonresponsive" the maintainers, so this could help me uncover more. I'd rerun the script but I'm on bad Internet connection.

### DIFF
--- a/portingdb/load_data.py
+++ b/portingdb/load_data.py
@@ -281,14 +281,17 @@ def load_from_directories(data, directories):
     for name, package in packages.items():
         releasever = None
         for rpm_name in package['rpms']:
-            nvr, sep, dist_tag = rpm_name.rpartition('.')
-            if dist_tag.startswith('fc'):
-                try:
-                    releasever = int(dist_tag[2:])
-                except ValueError:
-                    continue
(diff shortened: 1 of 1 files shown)


encukou:
It affects `qmtest` only.

``` console
$ python3 -c 'from portingdb.load_data import get_data; print("\n".join(n for n,p in get_data("data/")["packages"].items() if p["ftbfs_age"]))' > /tmp/a
$ git pull origin refs/pull/550/head
...
$ python3 -c 'from portingdb.load_data import get_data; print("\n".join(n for n,p in get_data("data/")["packages"].items() if p["ftbfs_age"]))' > /tmp/b
$ diff -U0 /tmp/[ab]
--- /tmp/a      2019-07-19 15:05:50.273686509 +0200
+++ /tmp/b      2019-07-19 15:06:04.960686306 +0200
@@ -118,0 +119 @@
+qmtest
```

PR #554, opened 2019-07-25
Two little dev usability fixes

encukou:
See commit messages

### DIFF
--- a/README
+++ b/README
@@ -18,7 +18,7 @@ To install it, first set up a
 [Python virtual environment](https://developer.fedoraproject.org/tech/languages/python/python-installation.html).
 Then run directly from Git working directory:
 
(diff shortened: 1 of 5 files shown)


hroncok:
Why was the second commit not needed before?

encukou:
I usually run the server first.

hroncok:
It still works for me without. What i the error you get? Do you have portingdb installed?

```
(venv) $ pip freeze | grep porting
-e git+git@github.com:fedora-python/portingdb.git@b25edf89461f616626c016c59c0af8579a7b9b72#egg=portingdb
```

Maybe we should docuement that instead?

encukou:
No, I don't have it installed.
Changed to install `-e.` instead of setting PYTHONPATH. It's no longer such a tiny change now :)

PR #556, opened 2019-07-30
3 improvements

encukou:
Best reviewed separately (but if there are no issues, let's let Travis handle them at once):
* Show non-Python dependents in package pages
* Adjust status descriptions that assumed this only is for Python 3 porting
* Move CSS to a separate file


portingdb/load_data.py
```diff
@@ -297,3 +297,13 @@ def load_from_directories(data, directories):
...
+
+    # Consolidate non-Python requirers
+    for name, package in packages.items():
+        nonpy_requirers = {}
+        package.setdefault('needs_py2_command', False)
+        for rpm in package['rpms'].values():
+            for kind, names in rpm.get('non_python_requirers', {}).items():
+               if names:
+                    nonpy_requirers.setdefault(kind, set()).update(names)
+        package['non_python_requirers'] = nonpy_requirers
```
hroncok: I've read this code 5 times and I have no idea what does it actually do. Sorry. I must miss some context.

hroncok:
Consider the *Adjust status descriptions that assumed this only is for Python 3 porting* and *Move CSS to a separate file* commits approved.

portingdb/load_data.py
hroncok: Oh, so we already track this in

https://github.com/fedora-python/portingdb/blob/4bce049022122493e4ce813bf20cf2224a85e0ad/dnf-plugins/py3query.py#L353-L356

portingdb/load_data.py
```diff
@@ -297,3 +297,13 @@ def load_from_directories(data, directories):
             package['ftbfs_age'] = max(CURRENT_FEDORA - releasever, 0)
         else:
             package['ftbfs_age'] = 0
+
+    # Consolidate non-Python requirers
+    for name, package in packages.items():
+        nonpy_requirers = {}
+        package.setdefault('needs_py2_command', False)
```
hroncok: I'Ve tried to remove this line and it seems the website works just fine without. Will check if there are any differences.

portingdb/load_data.py
encukou: Yes, but that's per RPM.
We probably don't need it per-RPM, but I'd like to keep old `fedora.json` formats loadable, so it's probably better to postprocess in load_data.

portingdb/load_data.py
encukou: You're right, that's a merge artifact.

hroncok:
```suggestion
```

hroncok:
As a followup readability improvement, maybe we can link pagure and spec with icons?

https://fontawesome.com/icons/code-branch?style=solid
https://fontawesome.com/icons/scroll?style=solid

encukou:
Would be that useful often? My first attempt at it looked less readable – too cluttered.
Note that portingdb uses Font Awesome 4, the previous version.

PR #557, opened 2019-07-31
Repurpose Groups for dependency analysis of (possible) FESCo exceptions

encukou:
IMO, the "Groups" are kind of useless right now.
This changes the groups to:
* Calibre
* Chromium browser
* GIMP
* Inkscape
* OLPC/Sugar
* Version Control & Trac

… and rewords the message, and makes a few tweaks to the way the info is shown.
I expect to do more changes as FESCo exceptions are filed.


hroncok:
FWIW pypy and pypy3 has exception so far (they don't actually bring in any python2 by default)

encukou:
Do you have a link to the exception?

hroncok:
RFE: Together N packages bringing in M Python 2 packages.

hroncok:
> Do you have a link to the exception?

https://pagure.io/fesco/issue/2130

encukou:
It should be "N packages with exception, bringing in M Python 2 packages." But the N might not be exact – this skips packages portingdb doesn't track,

encukou:
Added PyPy as preview.

portingdb/templates/index.html
```diff
@@ -154,9 +157,16 @@ <h2>Package Groups</h2>
                             >
                             {{ grp.name }}
                         </a>
+                        {% if 'exception' in grp %}
+                            <small>(<a href="{{ grp['exception'] }}">exc.</a>)</small>
+                        {% endif %}
```
hroncok: Suggestion: Else: (needs exc.)

encukou: Not necessarily; portingdb might be outdated. I think the warning above is enough.

encukou:
> It would be nice if the group page listed nontracked packages that are explicitly part of that group. 

Added. And I added links to exception *requests* like https://pagure.io/fesco/issue/2179 (repurposing `hidden`, which isn't a good name, but that should be OK).

encukou:
> Not sure if adding not approved exceptions is needed

IMO it's useful to see the dep tree even before an exception is requested. The link to the tracker is a possibly useful bonus.

> we can remove ptxdist if it gets rejected.

Yes; ptxdist is just the only example I had.

PR #562, opened 2019-08-06
Improve dependency trees, rename statuses, clarify this is Rawhide

encukou:
I nerd-sniped myself a bit last week on portingdb, but ran out of time.
These patches aren't as polished as I'd like, but I won't have much time for them, so putting them up for review. Merge any you like.

The first patch should make dependency trees on the package & group pages useful for scoping FESCo exceptions.

### DIFF
--- a/data/statuses.yaml
+++ b/data/statuses.yaml
@@ -51,7 +51,7 @@
 
     If this is a false positive, file a portingdb bug.
 - ident: idle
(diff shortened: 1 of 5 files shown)


hroncok:
The two smaller patches are good. The first one will require a more thorough review, but the output is indeed nicer.

PR #563, opened 2019-08-06
Keep track of old python- provides for python2- packages

hroncok:
This is the best I could come up with before a meeting. It appears to work as expected.

TODO:

- [x] Update docs.


encukou:
Pushed my notes as commits. Does this look good?

hroncok:
It does, thanks!

PR #564, opened 2019-08-07
Update status of edk2

frenzymadness:
edk2 contains multiple python subpackages but all of them depends only on Python 3 so no matter what purpose they serve, this is Py3-only package.


hroncok:
Restarting travis, something died there.

hroncok:
Thank You!

PR #566, opened 2019-08-10
Apply knowledge from Flock: Check uninstallable packages in seconds

hroncok:


hroncok:
Copy pasting stuff from other places. I wonder if it's time to create an utility library that can do stuff like:

 - see if there is a bugzilla tracked for a package
 - open a templated bugzilla fully automatically or in webbrowser, but with nice API
   -  possibly adding a comment to an existing one
 - determine package source package
 - give all packages from a source package
 - give latest builds from Koji tag / Copr repo with statuses
 - say if a package is retired
 - run `repoquery --{what,}{requires,provides}` over certain repo(s)
 - tell who is a package maintainer or give package maintainers for a package
 - ...

It seems that we do it again and again. [python-fedora](https://github.com/fedora-infra/python-fedora) unfortunately does nothing of this sort. I wish I had more time for this :(

encukou:
I don't think an utility library would be very maintainable...
Which of these would be best to write a library for?

hroncok:
> Which of these would be best to write a library for?

Bugzilla higher API. That's the bit that is copied all over the place and is nontrivial

PR #569, opened 2019-08-15
Update Fedora data

encukou:
Includes #568.
I might have run the old DNF command for the last update, so there are some extra naming status changes.

------

**idle** → **mispackaged** (1)
- icaro (/usr/bin/python)

**idle** → **missing** (1)
- scribus-generator

**idle** → **py3-only** (1) ♥
- redhat-lsb (not really ported)

**legacy-leaf** → **released** (1)
- python-enchant

**mispackaged** → **idle** (1)
- alacarte

Naming status changes
---------------------

**missing** → **requires misnamed** (3)
- geany-plugins
- fawkes
- kdepim3

**named correctly** → **missing** (1)
- scribus-generator

**requires ok** → **requires misnamed** (4)
- dmlite
- calibre
- pilas
- puddletag


hroncok:
> redhat-lsb (not really ported)

is there code to port?

encukou:
No, it just depends on /usr/bin/python and doesn't even really use it

PR #572, opened 2019-08-19
New bugs

encukou:
Remember the time I nerd-sniped myself on portingdb?
This is probably the last usable thing from that.

When portingdb started, it collected "links" for upstream URL, the py3 porting bug, Bugzillas, etc.
These are now usually unused – we don't want to store information in portingdb if we can avoid it.

This pulls bugs out of links, makes it possible to add more of them, tracks the trackers, and displays them on the package page:

image

### DIFF
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -32,6 +32,7 @@ BUGZILLA_URL = 'bugzilla.redhat.com'
 # Tracker bugs which are used to find all relevant package bugs
 TRACKER_BUGS = {
     1698500: "F31_PY2REMOVAL",
+    1625773: "PY2REMOVAL",
     1690439: "PY2FTBI",
     1312032: "PY3PATCH-AVAILABLE",
     1333770: "PY3PATCH-PUSH",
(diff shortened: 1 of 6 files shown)


hroncok:
This look useful, I recently thought: maybe we need to track the removal bugs here, but not mark packages mispackaged on them. Will check the code shortly!

dnf-plugins/py3query.py
```diff
@@ -30,22 +30,17 @@
...
-]
-# Trackers of bugs whose presence gives us additional information about the
-# status quo.
-ADDITIONAL_TRACKER_BUGS = [
-    1312032,  # PY3PATCH-AVAILABLE
-    1333770,  # PY3PATCH-PUSH
-    1432186,  # Missing PY3-EXECUTABLES
-]
+TRACKER_BUGS = {
+    1698500: "F31_PY2REMOVAL",
```
hroncok: and PY2REMOVAL, sadly.

hroncok:
```suggestion
    1625773: "PY2REMOVAL",
    1698500: "F31_PY2REMOVAL",
```

encukou:
Thanks! I already had a commit for adding it.
The data update is in #574.

encukou:
Merged via #574

PR #573, opened 2019-08-19
Update info on the Sugar stack

encukou:


hroncok:
The only benefit of the "dropped" status I see is to use it for "split" python2-foo packages. You said we don't want to track info here. But do we?

encukou:
I don't want to track info that is better put somewhere else. I don't know of a better place for "dropped".

"dropped" is for packages that will not be ported, but have a "py2-less" alternative packaged in Fedora (even if it's not a drop-in dependency).
Here I'm bending the rules a bit: even though `sugar-toolkit-gtk3` is still py2-only, it's the way to get things ported. So I think it's reasonable to say `sugar-toolkit` is dropped in favor of `sugar-toolkit-gtk3`.

PR #575, opened 2019-08-20
cairo-dock-plug-ins is fixed in rawhide

frenzymadness:
https://koji.fedoraproject.org/koji/buildinfo?buildID=1361142


encukou:
Indeed. Thanks!

PR #581, opened 2019-08-29
Update groups data

hroncok:
- add exception requests for postgresql and chromium
 - remove inkscape and calibre (Python 3 now)
 - trac-fedmsg-plugin was retired recently


encukou:
Thanks!
I'm also waiting for `python27` to add a group for Python 2 itself.

PR #582, opened 2019-08-29
Update Fedora data

encukou:
**idle** → **missing** (4)
- cloudtoserver
- cycle
- inkscape
- uniconvertor

**idle** → **py3-only** (2) ♥
- inkscape-psd (limb ∞)
- inksmoto (limb ∞)

**legacy-leaf** → **py3-only** (2)
- python-demjson
- python-regex

- gnome-feeds
- python-listparser
- scribus-generator
- solaar

**released** → **legacy-leaf** (1)
- python-scour

Naming status changes
---------------------

**named correctly** → **missing** (4)
- cloudtoserver
- cycle
- inkscape
- uniconvertor


hroncok:
I wonder why inkscape is not marked as py3 :confused:

hroncok:
```
$ repoquery --repo=koji{,-source} --requires inkscape | grep python
python-unversioned-command
python3-devel
python3-lxml
python3-numpy
python3-scour
```

python-unversioned-command should mark it as py, shouldn't it?

hroncok:
Got it, python-unversioned-command is buildtime only:

```
$ repoquery --repo=koji --requires inkscape | grep python
python3-lxml
python3-numpy
python3-scour
```

I'll make it depend on python3 as well.

hroncok:
https://src.fedoraproject.org/rpms/inkscape/pull-request/4

encukou:
Thanks!

PR #605, opened 2019-10-03
⏻ Add "orphaned" badges to package names

encukou:
image

I also had some code to show how long these were orphaned, but it takes too long to update (by querying pagure).

### DIFF
--- a/portingdb/templates/_base.html
+++ b/portingdb/templates/_base.html
@@ -46,7 +46,16 @@
         ignore_unversioned_requires=ignore_unversioned_requires,
     ) -}}
     {{- ftbfs_badge(pkg) -}}
+    {{- orphan_badge(pkg) -}}
 {%- endmacro %}
+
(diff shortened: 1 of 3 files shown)


hroncok:
Is the "upper index" position intentional?




I'd appreciate if the information was also presented on the detail page of the orphaned package. Maybe even more verbosely. Byt this is not a blocker.

encukou:
> Is the "upper index" position intentional?

Yes*.

<i>*⁾ I wanted this to look like a note.</i>

> I'd appreciate if the information was also presented on the detail page of the orphaned package. 

Added.

hroncok:
I'll see if I can publish the packages/weeks from https://churchyard.fedorapeople.org/orphans.txt as json.

hroncok:
See https://churchyard.fedorapeople.org/orphans.json

encukou:
See https://github.com/fedora-python/portingdb/pull/611 :)

PR #609, opened 2019-10-04
python-openid is Python 3 only

hroncok:


hroncok:
Oh, a mistake in the package name.

PR #611, opened 2019-10-04
Add hourglasses to orphan packages

encukou:
image


encukou:
Hmm, Python 3.7 added the convenient `datetime.fromisoformat`. I don't think requiring 3.7+ should be any trouble nowadays?

portingdb/load_data.py
```diff
@@ -330,3 +330,8 @@ def load_from_directories(data, directories):
                if names:
                     nonpy_requirers.setdefault(kind, set()).update(names)
         package['non_python_requirers'] = nonpy_requirers
+
+    for name, isodt in data_from_file(directories, 'orphans').items():
+        if name in packages:
+            last_change = datetime.datetime.fromisoformat(isodt)
+            packages[name]['last_repo_metadata_change'] = last_change
```
hroncok: In case of else, you can assume the package is orphaned but not yet in orphans.json, hence it was orphaned "now".
(Assuming the property is only added to packages that are orphaned, or at least ignored otherwise)

the only other way I can think of is that it was already retired and the repo mirror used to update portingdb doesn't know it yet, but the one generating the orphans.json knew it. I find it highly unlikely (orphans.json takes copule hours to be published) and the data for the retired package would be wrong anyway.

encukou: The else is mainly for non-Python packages.

But in any of the 3 cases (mine & 2 yours), we should be fine having no data and showing no hourglass, right?

hroncok: right

hroncok:
requiring 3.7+ is OK, should probably be mentioned in the README

portingdb/htmlreport.py
```diff
@@ -269,13 +269,13 @@ def format_time_ago(date):
         else:
             return "a year ago"
     # Months
-    elif diff.days >= 31:
+    elif diff.days >= 8*7:
         if diff.days >= 2 * 30:
             return "{} months ago".format(math.floor(diff.days / 30))
         else:
             return "a month ago"
```
hroncok: the first elif means >= 56 days, the second one is >= 60. There is an else for "a month ago" for 4 days only. Is that worth it? It's almost 2 months anyway at that point. If we change `8*7` to `9*7`, we can get rid of the inner if.

portingdb/htmlreport.py
```diff
@@ -269,13 +269,13 @@ def format_time_ago(date):
...
     # Weeks
-    elif diff.days >= 7:
+    elif diff.days >= 10:
```
hroncok: what was the reason for this change?

portingdb/templates/package.html
```diff
@@ -107,6 +107,22 @@ <h1>Package {{ pkg.name }} </h1>
...
+                            datetime="{{ pkg['last_repo_metadata_change']}}"
+                            title="{{ pkg['last_repo_metadata_change']}}"
+                        >
+                            {{ pkg['last_repo_metadata_change'] | format_time_ago -}}
+                        </time>.
+                        Check the
+                        <a href="https://churchyard.fedorapeople.org/orphans.txt">
+                            orphan report
+                        </a>
+                        for details
```
hroncok:
```suggestion
                        for details.
```
Sorry for the late notice, just found out about this now.

hroncok:
All is good in the world.

PR #612, opened 2019-10-04
Document the recent purpose of this tool

hroncok:
We don't port to Python 3, we kill Python 2.


encukou:
IMO the point is to find out what needs to be kept (and ported in the future)?
Anyway, that matches what your change says.

PR #619, opened 2019-10-22
Update taskotron-python-versions

hroncok:
It contains various new packages in NAME_NOTS

Mostly doing this to let @encukou know to update his venv.


encukou:
Thanks, updated!

PR #620, opened 2019-10-22
Track new approved FESCo exceptions for autodownloader and mozjs60

hroncok:
autodownloader: https://pagure.io/fesco/issue/2248
mozjs60: https://pagure.io/fesco/issue/2249


encukou:
Thanks!

PR #628, opened 2019-11-05
Show icons for packages with FESCo exceptions

encukou:
image

---

image

---

image

---

image


hroncok:
I was actually hoping to get something like this. Merge at will if it work, it looks awesome on the pictures.

Issue #633, opened 2019-11-12
packages in "needed for" list lack the exception symbol


hroncok:
![_](https://user-images.githubusercontent.com/2401856/68631830-8649d500-04ec-11ea-8d0b-d7218ba938b0.png)

Some of the listed packages here have exceptions, but it is not visible just by looking at them here.

hroncok:
the reverse relation has the same issue:

![_](https://user-images.githubusercontent.com/2401856/68631978-299aea00-04ed-11ea-8539-b909c112e2ec.png)

encukou:
I think it would add too much clutter. Do you think it would be useful?
![image](https://user-images.githubusercontent.com/302922/68654435-34c23a00-052e-11ea-9601-4499f612c71c.png)

These sections aren't very useful for removing py2, IMO. They only show one side of the graph. What are you using them for?

PR #634, opened 2019-11-12
index.html: Add exception icons in the dependencies/dependents

encukou:
Fixes: https://github.com/fedora-python/portingdb/issues/633


hroncok:
I don't want to force this on you in case it feels liek too much information squeezed into a small place.

encukou:
You probably use this list more than I do :)

hroncok:
Thanks!

PR #635, opened 2019-11-12
Update groups according to FESCo exception requests

encukou:


data/groups.yaml
```diff
@@ -11,18 +11,51 @@ gimp:
...
     - pycairo
 olpc:
-    name: OLPC/Sugar
+    name: Sugar desktop environment
+    exception: https://pagure.io/fesco/issue/2267
     hidden: true
     packages:
-    # dnf repoquery --repo rawhide 'sugar-*' 'olpc-*' --qf '- %{name}'
     - olpc-kbdshim
-    - olpc-netutils
```
hroncok: I think this should remain here.It is a Python 2 OLPC package, it just has x86_64 excluded, so it sometimes doesn't show up.

data/groups.yaml
```diff
@@ -176,16 +206,35 @@ mozjs60:
...
     - gr-fcdproplus
     - gr-hpsdr
-    - gr-iio
     - gr-iqbal
     - gr-osmosdr
     - gr-rds
+    - python-zmq
+
+    # gr-iio, gnuradio + its immediate dependencies + their runtime deps
+    # (These aren't listed in the exception request at time of writing, but
```
hroncok: neither is python-zmq

data/groups.yaml
```diff
@@ -176,16 +206,35 @@ mozjs60:
...
+    - python-lxml
+    - python-mako
+    - python-markupsafe
+    - python-pyopengl
+    - python2-numpy
+    - python-six
+    - pygtk2
+    - pycairo
+    - pygobject2
+    - wxPython
```
hroncok: this is now retired

data/groups.yaml
```diff
@@ -176,16 +206,35 @@ mozjs60:
     - mozjs60
 gnuradio:
     name: GNU Radio
+    exception: https://pagure.io/fesco/issue/2272
```
hroncok: Closed, all ported. Give @yarda a medal.

encukou: Wow! What great news!

This comment appeared to me after I pushed a commit to fixup the group. I'll amend it.

encukou: I'll remove the group when it disappears in the compose.

data/groups.yaml
```diff
@@ -11,18 +11,52 @@ gimp:
...
+    - pybox2d
+    - python-BeautifulSoup
+    - gnome-python2
+    - gnome-python2-desktop
+    - pygtk2
+    - python-docutils
+    - python-dateutil
+    - python-six
+    - python-simplejson
+    - python-nose
```
hroncok:
```suggestion
```

PR #641, opened 2019-11-18
Record exception request for git-remote-hg

encukou:


hroncok:
Given the way we track this, should mercurial be also listed?

encukou:
I don't think the exception covers Mercurial. As I see it, when the Mercurial exception is gone, git-remote-hg won't keep python2-mercurial in.

PR #644, opened 2019-11-21
Add new Bugzilla trackers

hroncok:


encukou:
Could we also remove some old ones?

hroncok:
Done.

encukou:
Thank you!

PR #646, opened 2019-11-25
Update Fedora data

encukou:
**idle** → **missing** (6)
- nfspy
- pypoppler
- python-arc
- python-subprocess32
- virtio-forwarder
- yaws

**idle** → **py3-only** (1) ♥
- votca-xtp (junghans 2→3)

**legacy-leaf** → **py3-only** (2)
- python-attrs
- python-lxml

**missing** → **py3-only** (2)
- koan
- oval-graph

**released** → **legacy-leaf** (1)
- python-zmq

**released** → **py3-only** (3)
- PyQt4
- protobuf
- python-musicbrainzngs

Naming status changes
---------------------

**misnamed** → **named correctly** (1)
- PyQt4

**named correctly** → **missing** (6)
- nfspy
- pypoppler
- python-arc
- python-subprocess32
- virtio-forwarder
- yaws

**requires misnamed** → **missing** (1)
- nextcloud-client

**requires misnamed** → **requires ok** (1)
- ioprocess


data/orphans.json
```diff
@@ -96,7 +100,7 @@
     "jogl2": "2019-11-13T20:01:15",
     "jove": "2019-11-13T20:01:16",
     "lcms": "2019-11-05T15:31:46",
-    "libee": "2019-11-07T16:16:54",
+    "libee": "2019-11-23T07:08:04",
```
hroncok: What? :confused:

encukou: Some owner change happened?

hroncok: Probably. But that would need to be from orphan to non-orphan to orphan. And there was no releng ticket to unrophan libee. I guess what the timestamp actually tracks.

PR #653, opened 2020-01-09
Update Fedora data

encukou:
**idle** → **missing** (15)
- configsnap
- exaile
- fslint
- gdesklets
- getmail
- glue-validator
- gourmet
- k3d
- magicor
- pagekite
- rocket-depot
- seahorse-adventures
- shedskin
- squeal
- sx

**missing** → **py3-only** (2)
- gnulib
- python-wloc

**released** → **py3-only** (1)
- pygame

Naming status changes
---------------------

**named correctly** → **missing** (15)
- configsnap
- exaile
- fslint
- gdesklets
- getmail
- glue-validator
- gourmet
- k3d
- magicor
- pagekite
- rocket-depot
- seahorse-adventures
- shedskin
- squeal
- sx


encukou:
> There are some leftver updates, but so they were before and we might get some unretirements anyway, so we should keep them for at least 8 weeks.

Sorry, I can't parse this. Could you reword it?

hroncok:
Some removed packages have overrides defined in [fedora-update.yaml](https://github.com/fedora-python/portingdb/blob/master/data/fedora-update.yaml). We keep them there on purpose, in case they get unretired. After 8 weeks, they cannot be unretired without a re-review, so we can safely purge them after that time.

encukou:
I see. Thanks for explaining!
Actually at this point, we can probably get away with not changing fedora-update until portingdb EOL :)

PR #658, opened 2020-01-17
Fix offlineimap exception link

hroncok:


encukou:
Oops! Thanks for catching this!

PR #668, opened 2020-03-02
Update Fedora data; record exception requests

encukou:
**idle** → **missing** (4)
- python2-typing (retired)
- slimdata (BR fix – builds without Python)
- sugar-connect (retired)
- tuxpaint (BR fix – builds without Python)

**idle** → **py3-only** (4) ♥
- libosmocore (BR fix)
- openclonk (BR fix – removed docs)
- openmsx (BR fix – py3 build backport)
- tuxpaint-stamps (BR fix)

**missing** → **idle** (1)
- pspp (dependency uncovered after FTBFS fix; filed [BZ#1809024](https://bugzilla.redhat.com/show_bug.cgi?id=1809024))

- nanovna-saver
- nvml
- sourcextractor++
- wsdd

**py3-only** → **missing** (2)
- antimony (retired)
- mint-themes (BR removal – bundless CSS rather than rebuilding it)


hroncok:
The first commit has wrong keys, the second one fixes it. Not sure if we care enough, but it's certainly weird.

encukou:
Oops, you're right!
Rebased.

PR #676, opened 2020-04-07
Embed functions from taskotron-python-versions

hroncok:
This is a preparation for EOLing taskotron-python-versions.


encukou:
Since we're touching this, I'd rather remove the unneeded functionality than keep carrying it. See https://github.com/fedora-python/portingdb/pull/677

PR #677, opened 2020-04-08
Remove or or embed functions from taskotron-python-versions

encukou:
Remove the checks for naming policy (is_misnamed and unversioned_requirers).
That part is done; if new packages named python-X appear they
better be python3 (which is not cool according to guidelines,
but not a problem worth tracking).

Remove special-casing NAME_NOTS. This means packages like
python-rpm-generators and python-rpm-macros start showing up as
build dependencies (in python3-only green).

The `have_binaries` function from taskotron-python-versions is embedded.

This is preparation for EOLing taskotron-python-versions.

Closes: https://github.com/fedora-python/portingdb/pull/676

### DIFF
--- a/dnf-plugins/py3query.py
+++ b/dnf-plugins/py3query.py
@@ -23,10 +23,6 @@ from dnfpluginscore import _
 import bugzilla  # python-bugzilla
 import yaml
 
-from taskotron_python_versions.executables import have_binaries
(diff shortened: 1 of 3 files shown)


encukou:
@hroncok: I like the data this generates. Does the patch look sane?

encukou:
Thanks!

PR #682, opened 2020-06-05
Adapt for new orphans.json format

hroncok:
The new format is not there yet, will update here when ready.


hroncok:
The new format is up.

encukou:
Thank you!

PR #698, opened 2021-02-16
Update Fedora data

encukou:
99.0%!

I've also removed groups that are already ported.

**idle** → **missing** (4)
- sugar-getiabooks
- sugar-infoslicer
- sugar-ruler
- vte

**idle** → **py3-only** (2) ♥
- audacity (limb 26→27)
- git-cinnabar (qulogic 3→4)

**missing** → **py3-only** (6)
- oc-inject
- python-imageio
- python-prefixed
- python-pynn
- python-sphinx-tabs
- python-wled

**py3-only** → **missing** (1)
- gnatcoll-db

**released** → **py3-only** (1)
- python-rpm-macros


data/groups.yaml
```diff
@@ -172,22 +174,11 @@ postgresql:
     exception: https://pagure.io/fesco/issue/2217
     packages:
     - postgresql
-pypy:
```
hroncok, line 175: pypy is not ported, it is just that currently, all the builds were done with pypy. But when bootstrapping, they are done with python2. Not sure if we want to keep this here or not, just saying how it is.

hroncok (approved):

data/groups.yaml
encukou: I'll put it back.It's not something portingdb tracks well, but I guess it deserves a comment.

hroncok:
See also https://github.com/fedora-python/fedoralovespython.org/pull/58

Issue #705, opened 2021-07-19
Upcoming Bugzilla API change


encukou:
It seems an API change is coming to RH Bugzilla: large results will start to be paginated.
Since I don't use portingdb bug links too much, I intend to spend about an hour adapting to this change (after it goes live), and remove bug links from portingdb if I don't make it.
Let me know if I should put in more effort.

hroncok:
Nah, the bugzillas here are not used any more, if it is too much effort, we can drop them.

hroncok:
Petr, where can I find more info about this?

encukou:
The change is now live and I see it already hit you elsewhere: https://github.com/python-bugzilla/python-bugzilla/issues/149
I heard about the change in Red Hat internal lists; unfortunately I don't know about public announcements.

PR #706, opened 2021-08-03
Make smaller bugzilla queries to workaround pagination

hroncok:
Fixes https://github.com/fedora-python/portingdb/issues/705

Untested.


hroncok:
Included in https://github.com/fedora-python/portingdb/pull/707

PR #708, opened 2021-08-03
Switch to GitHub Actions

encukou:
gh-pages currently isn't building; Travis CI stopped working. (This could be fixed, but their free plan is very limited now.)


encukou:
Tested in my fork, see https://github.com/encukou/portingdb/runs/3230635688 & https://github.com/encukou/portingdb/tree/gh-pages

hroncok (approved):

PR #719, opened 2022-06-30
Add Pyhon 3.10 and 3.11 to SEED_PACKAGES

hroncok:
Not sure why this exists at all, it seemed to work even without that :/


encukou:
Oh, it finds a few more stragglers:
    
    - hpx
    - mingw-Cython
    - mingw-numpy
    - python3.10

PR #722, opened 2022-11-14
Update Fedora data

encukou:
**idle** → **missing** (1)
- slv2

**missing** → **py3-only** (57)
- android-tools
- ardour7
- crosswords
- crosswords-puzzle-sets-puzzlepull
- dnf5
- flamegraph
- libipuz
- libloc
- mingw-python-build
- mingw-python-flit-core
- mingw-python-installer
- mingw-python-pep517
- mingw-python-pip
- mingw-python-wheel
- mrack
- nagios-plugins-systemd
- packetdrill
- piccata
- python-adjustText
- python-azure-mgmt-quota
- python-colcon-alias
- python-colcon-installed-package-information
- python-django-mailman3
- python-dunamai
- python-fastapi
- python-formulaic
- python-icoextract
- python-interface-meta
- python-ipuz
- python-mistune08
- python-mizani
- python-oslo-metrics
- python-palettable
- python-pdm-pep517
- python-pidfile
- python-plotnine
- python-poetry-dynamic-versioning
- python-poetry-plugin-export
- python-puzpy
- python-pyclip
- python-pytz-deprecation-shim
- python-qemu-qmp
- python-robotframework
- python-rpi-gpio2
- python-rust-update-set
- python-scikit-misc
- python-setupmeta
- python-sphinx-typlog-theme
- python-syslog-rfc5424-formatter
- python-tpm2-pytss
- python-xattr
- rEFInd
- rnp
- roc-toolkit
- rpmdistro-repoquery
- tacacs
- vcs-diff-lint

**py3-only** → **missing** (16)
- blueberry
- mingw-python-setuptools_scm
- python-APScheduler
- python-coreapi
- python-coreschema
- python-drf-yasg
- python-itypes
- python-krbcontext
- python-pacpy
- python-pecan-notario
- python-phabricator
- python-pydenticon
- python-pytest-capturelog
- python-qrencode
- scudcloud
- sgmanager


hroncok (approved):

CAM-Gerlach:
A few notes from a longtime follower:

* The [`slv2` issue](https://bugzilla.redhat.com/show_bug.cgi?id=1807546) appears to not have been closed, despite the package being [retired](https://src.fedoraproject.org/rpms/slv2/c/b42d049a5558569e2b075848e32b6a22bcbe4a41?branch=rawhide) (assuming I'm interpreting that right). Is that intended?
* `flowcanvas` is long-retired upstream and is not required by anything else, so presumably it should be orphaned/removed per you two's last comment on [its issue](https://bugzilla.redhat.com/show_bug.cgi?id=1805233), but it was never actually orphaned/removed.
* Per you, `kdissert` was renamed to `Semantik` upstream and has long since been ported to Python3, but the separate Fedora package for the latter was retired. Seems like it should be orphaned and removed, possibly with adding `Semantik` as a replacement
* The other trac plugins covered by a blanket Trac FESCO exception were orphaned and removed a while ago, but [email2trac](https://bugzilla.redhat.com/show_bug.cgi?id=1738907) was not, despite there similarly being no response from the maintainer. A user recently commented that the upstream seems to have moved to GitLab, but nevertheless it still does not appear to have Python 3 support and not too much activity, at least as far as I can tell. Perhaps it should be handled the same way its brethren previously were, with an orphan/removal?
* NFSTest had a previous FESCO exception years ago, but there appears to be no activity [on the issue](https://bugzilla.redhat.com/show_bug.cgi?id=1731377) since then, so perhaps it should be pinged again or orphaned? The [upstream](git.linux-nfs.org/?p=mora/nfstest.git) still appears to be active, but it is unclear if it is Python 3 compatible.

encukou:
@CAM-Gerlach Yeah, we (Red Hat python-maint team) are just tracking stuff right now. The next push for porting/retiring stuff will probably happen after Gimp 3.0 is out, or when Python 2 is retired in RHEL.
We don't actually maintain the “idle” packages, so we can either ask the maintainer nicely, or use a [process](https://docs.fedoraproject.org/en-US/fesco/Policy_for_nonresponsive_package_maintainers/) to claim that the maintainer is unresponsive and the package should be removed (or transferred to us).  Not really worth it if it doesn't actually unblock removal of another unwanted package.

For slv2: bugs don't generally auto-close when a package is retired. It's only retired in Rawhide, but the issue still exists in older Fedoras. Also, the package could be un-retired.

hroncok:
> * The [`slv2` issue](https://bugzilla.redhat.com/show_bug.cgi?id=1807546) appears to not have been closed, despite the package being [retired](https://src.fedoraproject.org/rpms/slv2/c/b42d049a5558569e2b075848e32b6a22bcbe4a41?branch=rawhide) (assuming I'm interpreting that right). Is that intended?

I've closed the bugzilla, unretirement is unlikely and if it happens, we can always reopen it.

